PR #4, opened 2019-12-03
C# part 2 and Ruby part 1 and 2 stubs

ErikSchierboom:


iHiD:
Sweet. I'll try and get this onto the webservers now.

PR #7, opened 2019-12-03
python: exercises for part 1

yawpitch:
Adds exercises for Python part 1 and boilerplate for part 2.


exercises/python-1-a/README.md
```diff
@@ -0,0 +1,23 @@
...
+This exercise should take 10-15 minutes.
+There is no right or wrong way to approach it - just do what feels the most natural to you.
+Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.
+
+# Instructions
+
+Your task is to compress a UTF-8 [`str`][str] to [`bytes`][bytes] using [Run Length Encoding (RLE)][rle], a relatively straightforward compression scheme that takes any continuous run of 1 or more of the same byte and compresses it into two bytes, the first being the length of the run and the second being the byte itself.
+
+We expect you to define a function `compress` that takes a single argument which is a UTF-8 `str` and returns a single even-length `bytes` value. Since individual UTF-8 characters can be more than one byte long you should encode the `str` to `bytes` _before_ applying the RLE compression. For the purposes of this task you can assume that no run will ever be more than 255 bytes long, , so no run will ever require more than one byte to encode.
```
ErikSchierboom:
```suggestion
We expect you to define a function `compress` that takes a single argument which is a UTF-8 `str` and returns a single even-length `bytes` value. Since individual UTF-8 characters can be more than one byte long you should encode the `str` to `bytes` _before_ applying the RLE compression. For the purposes of this task you can assume that no run will ever be more than 255 bytes long, so no run will ever require more than one byte to encode.
```

exercises/python-1-a/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+For example:
+
+```python
+>>> encode("abbcccddddaaaaaa")
+b'\x01a\x02b\x03c\x04d\x05e'
```
ErikSchierboom:
```suggestion
b'\x01a\x02b\x03c\x04d\x05a'
```

yawpitch: Good spot, thanks!

exercises/python-1-a/README.md
yawpitch: Ahh, the double comma ... fixed.

ErikSchierboom:
@iHiD Could you take a look too?

exercises/python-1-a/README.md
```diff
@@ -0,0 +1,23 @@
+# Introduction
+
+This is Part 1 of our research into how people write Python differently.
```
iHiD: Can we standardise on this sentence-per-line README approach everywhere if we do it here pls?

ErikSchierboom: This renders in the same way as the ones that have no newlines. My personal preference is to keep things on the same line instead of these (ignored) newlines, to have it visually conform to what is actually rendered.

exercises/python-1-b/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```python
+>>> transmit("Hello!")
+"HOTEL ECHO LIMA LIMA OSCAR"
+
+>>> transmit("NCC-1701-D")
+"NOVEMBER CHARLIE CHARLIE ONE SEVEN ZERO ONE DELTA"
+```
+
+Next, we expect you to define a function `receive` that will take a NATO code word transmission and convert it back to a readable message. Because the phonetic alphabet is a lossy transmission format whitespace will be missing, and to avoid ambiguity all letters will be in uppercase.
```
iHiD: > whitespace will be missing

Am I right in understanding this means that you might get `HOTEL ECHOLIMA LIMAOSCAR` as input? If so, could we update one of the examples below with that? If I've misinterpreted can we update the text above to clarify :)

iHiD: Oh, does it mean that the original whitespace will be missing?

exercises/python-1-a/README.md
iHiD: Yeah - I think we need to decide on an Exercism standard and use it throughout the project. This PR almost certainly isn't the place for that though, so let's not worry about it here. I'll speak to Katrina and see if she has a hard preference, and if not, we can open it for a fun discussion! :)

iHiD:
Awesome. Thanks. Two small comments for clarity (one of which you can ignore). I'm interested in seeing where these go for the part 2s!

exercises/python-1-a/README.md
sshine: Very good to have standards. I noticed that GitHub's margin is <80 characters, which may be worth taking into account when making such a Markdown formatting standard.

exercises/python-1-b/README.md
yawpitch: Good point, I'll clarify that one.

exercises/python-1-a/README.md
yawpitch: I've made the newlines consistent pending an outcome on this "fun" discussion.

exercises/python-1-b/README.md
yawpitch: @iHiD are the new wording and examples more clear?

exercises/python-1-b/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```python
+>>> transmit("Hello, World!")
+"HOTEL ECHO LIMA LIMA OSCAR WHISKEY OSCAR ROMEO LIMA DELTA"
+
+>>> transmit("NCC-1701-D")
+"NOVEMBER CHARLIE CHARLIE ONE SEVEN ZERO ONE DELTA"
+```
+
+Next, we expect you to define a function `receive` that will take a NATO code word transmission formatted as described above and convert it back to a readable message. Because the phonetic alphabet is a lossy transmission format only the letters and digits from the original message will be included in the output. To avoid introducing ambiguity all whitespace will be stripped from the output and letters will be in uppercase.
```
iHiD: @yawpitch So this is definitely clearer, yeah. The only bit that still confuses me a bit is "To avoid introducing ambiguity", as in my mind it's more of an essential thing (ie we don't know where the whitespace is rather than a decision we're making). I think we could just remove those four words and it would become clearer for me.

yawpitch: The previous version was "to avoid ambiguity all letters will be in uppercase" ... the whitespace loss is, indeed, essential, but the choice of final case isn't exactly. Hearing the spoken words they might be recorded in any case ... an uninitiated receiver might title case the proper nouns ("Charlie", "India", "Lima", "Mike", "November", "Oscar", etc) and lowercase the rest ... it's not inconceivable that the cultural norms of any two such receivers might result in the same message having different semantic implications depending on case choice. It becomes essential then to choose a single casing, but because you'd nominally be receiving the transmission over audio transmission it's not essential that either upper or lower case is the casing chosen.

Still, at this point I agree with you, removing "to avoid introducing ambiguity" removes ambiguity, so I'll push that shortly.

iHiD:
@yawpitch Want to rebase this then I can merge it and you can start testing?

yawpitch:
@iHiD rebased and pushed ... BTW the config.json on master was missing a command and was invalid.

PR #10, opened 2019-12-04
haskell-1-a: Copy csharp-1-a experiment

sshine:


sshine:
Ah, I didn't know.

I don't object to this, but why?

sshine:
I do enjoy making up exercises, but I don't know what hypothesis to test.

Reading [Announcing Exercism Research](https://exercism.io/blog/announcing-exercism-research), I'm still not sure what hypothesis I would be testing for Haskell, so I figured: You probably had already put some thought into this exercise, so whatever you'd be testing, this one would test the same.

Note that I've pushed haskell-1-b here as well, since this was done before discussing that experiments cannot copy the base problem. I'll close this issue for now.

ErikSchierboom:
> I don't object to this, but why?

> Reading Announcing Exercism Research, I'm still not sure what hypothesis I would be testing for Haskell, so I figured: You probably had already put some thought into this exercise, so whatever you'd be testing, this one would test the same.

Maybe @iHiD could shed some more light on this?

iHiD:
Ah sure. So we're presuming that people will complete exercises in multiple languages. If they solve the same exercise in Haskell then C#, their mental reasoning in Haskell will "leak" into how they solve C#, making the results less accurate.

You could make a clone of the exercise with a few small changes that mean it doesn't feel familiar to the person, but follows the same sort of rationale.

One other thing I would say though is that this exercise was designed with the thought of how different developers might approach it using C#. If only one of those ways was valid/doable in Haskell, the exercise might not throw up other variants that we want to observe.

sshine:
> You could make a clone of the exercise with a few small changes

This sounds like a good plan. I'll make some twists.

> this exercise was designed with the thought of how different developers might approach it using C#. If only one of those ways was valid/doable in Haskell, the exercise might not throw up other variants that we want to observe.

Do you think there is only one valid/doable way to write code in Haskell?
(I'm asking out of curiosity as this may be a learning bias.)

I saw that @ErikSchierboom posted multiple different C# strategies in Slack but did not commit more than one example solution. Should I seek out and document the different ways that people might solve this exercise in Haskell? (Explicit recursion, folding, zipping, sorting, regex...)

ErikSchierboom:
@sshine I'll be fixing this in a PR! I should have included them in the repository.

sshine:
> I should have included them in the repository.

I will do the same then.

I look forward to seeing how you intend to organize multiple example solutions.

iHiD:
> Do you think there is only one valid/doable way to write code in Haskell?

@sshine I don't speak Haskell, so I have no idea :) 

What I really meant was that an exercise might be designed to be be solved really neatly in a functional or object-oriented way - ie there is an *obvious* FP solution and an *obvious* OOP solution, and we're interested in seeing which one someone takes. For a language that doesn't support OOP the exercise might then be not be well designed. It might still be solvable in different ways (e.g. recursion, folding, zipping, sorting, regex), but as it wasn't designed to be solved in those ways there might be a more obvious approach that people would choose to take, regardless of their backgrounds. Whereas a specific FP-designed exercise might offer multiple approaches as optimal, and therefore we might get more from it.

sshine:
> an exercise might be designed to be be solved really neatly in a functional or object-oriented way - ie there is an obvious FP solution and an obvious OOP solution, and we're interested in seeing which one someone takes.

OK, so this clarifies the answer to a question I asked yesterday about what hypothesis we're testing. So naturally we only look for FP solutions in Haskell, so a Haskell experiment exercise would never be about exploring FP vs. OO.

> For a language that doesn't support OOP the exercise might then be not be well designed.

How does one decide?

> It might still be solvable in different ways (e.g. recursion, folding, zipping, sorting, regex), but as it wasn't designed to be solved in those ways *[...]*

I didn't design the Haskell variations of `csharp-1-a` and `csharp-1-b` to be solved in any particular way.

> *[...]* there might be a more obvious approach that people would choose to take, regardless of their backgrounds.

So I agree that without thinking "why am I making this experiment?" I'm not actually testing anything, I'm just making fun exercises. This is why I copied `csharp-1-a` and `csharp-1-b` because I thought "Whatever undocumented hypothesis, we'll test that in Haskell, too."

Since the subject of `csharp-1-a` is strings, and the subject of `csharp-1-b` is queues, I figured that solutions to these exercises would yield some variants that are very typical of the Haskell track. For example, one way to assess progression we see on the Haskell track:

- Explicit recursion
- Folding
- Specialized higher-order functions (e.g. `Map.fromListWith` or `Set.unions`)

Another way to assess progression is brevity:

- Explicitly recursive functions typically don't reuse much.
- Folding typically lets one abstract out parts of domain logic, but due to the asymmetry of a fold, the folding function most often needs to be specialized.
- Specialized recursion gives way to pointfree notation and heavy use of existing combinators. (A part of Haskell-fu is writing functions that are highly composable. For example, while OCaml might have `fold_until`, Haskell's non-strictness means that you do something like `takeWhile`/`iterate`, `foldMap` or `foldM`.)

If the aim is to uncover how solutions made by people with different backgrounds can be placed on such scales, any problem will yield these aspects including the present `haskell-1-a` and `haskell-1-b`. (See for example Haskell mentoring notes for [`leap`](https://github.com/exercism/website-copy/blob/master/tracks/haskell/exercises/leap/mentoring.md) and [`nucleotide-count`](https://github.com/exercism/website-copy/blob/master/tracks/haskell/exercises/nucleotide-count/mentoring.md) for a similar attempt to uncover basic strategies.)

> Whereas a specific FP-designed exercise might offer multiple approaches as optimal, and therefore we might get more from it.

That sounds very true.

I'll have to think about specific FP-designed exercises based on curriculum. But I might not have a full language-concepts overview ready before January 2020. So until then, I'd like to work with the assumption that an experiment with an arbitrary problem and an undocumented hypothesis can show results about how people code Haskell, even though Haskell doesn't have OO.

ErikSchierboom:
> For example, one way to assess progression we see on the Haskell track:
> 
> - Explicit recursion
> - Folding
> - Specialized higher-order functions (e.g. Map.fromListWith or Set.unions)

This is definitely one aspect in which I'm very interested to see differences between students. I've done some digging through F# data and found some correllation between already knowing a functional language and using specialized higher-order functions. I would expect students with less functional experience to sooner use explicit recursion.

> Another way to assess progression is brevity:

> - Explicitly recursive functions typically don't reuse much.
> - Folding typically lets one abstract out parts of domain logic, but due to the asymmetry of a fold, the folding function most often needs to be specialized.
> - Specialized recursion gives way to pointfree notation and heavy use of existing combinators. (A part of Haskell-fu is writing functions that are highly composable. For example, while OCaml might have fold_until, Haskell's non-strictness means that you do something like takeWhile/iterate, foldMap or foldM.)

Also all very interesting aspects!

iHiD:
> Since the subject of csharp-1-a is strings, and the subject of csharp-1-b is queues, I figured that solutions to these exercises would yield some variants that are very typical of the Haskell track.

@sshine I think that's the key thing, yes. If they do this, then they're good target exercises. You're the best placed to make that judgement call. If you're happy that they are, then just varying them slightly would be enough I think.

sshine:
> varying them slightly would be enough I think

Since the `haskell-1-a` and `haskell-1-b` of this PR are basically the same exercise with a few twists:

```haskell
wordsWithMostVowels =
  maximumBy (comparing vowelCount) . words
  where vowelCount = length . filter isVowel

quickestQueue = minimumBy (comparing waiting)
  where waiting (Queue speed length) = speed * length
```

So I think it's basically harder to make an interesting problem that is a slight variation rather than a substantial variation. For example, keeping with the theme of the exercise seems more of a constraint than a relief.

I've changed one of them into "Find the recipe that requires you to buy as few ingredients you don't already have" (it's still more or less the same exercise) and I haven't come up with the other one, but probably something related to queues that does not have the same problem structure.

```haskell
chooseRecipe :: Set Ingredient -> [Recipe] -> Recipe
chooseRecipe ingredients = minimumBy (comparing missing)
  where missing = length . (\\ ingredients)
```

ErikSchierboom:
That sounds like a great idea.

iHiD:
I like this too :)

PR #14, opened 2019-12-04
csharp: add test messages for test results

ErikSchierboom:
@iHiD Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # Exercism Research: Experiment 1 Track
 
-[![Build Status](https://travis-ci.org/exercism/research_experiment_1.svg?branch=master)](https://travis-ci.org/exercism/research_experiment_1)
+[![Build Status](https://travis-ci.org/exercism/research_experiment_1.svg?branch=master)][travis-badge]
 
 This is a language track which holds the exercises for Exercism Research's first experiment.
 
 ## Exercises
 
-All exercises in this track will be bonus exercises. This means that their `core` property will be set to `false` and the `unlocked_by` property set to `null` in the [`config.json` file](https://github.com/exercism/research_experiment_1/blob/master/config.json). Here's an [example exercise entry](https://github.com/exercism/research_experiment_1/blob/master/config.json#L10).
+All exercises in this track will be bonus exercises. This means that their `core` property will be set to `false` and the `unlocked_by` property set to `null` in the [`config.json` file][config-json-file]. Here's an [example exercise entry][config-json-exercise].
 
 Each exercise must use the following naming format: `<track>-[1|2]-[a|b]`. The number part indicates which part of the exercise it represents (first part is start from scratch; second part is start with existing code). The letter part indicates which of the two exercises it is (remember: each track should have two exercises).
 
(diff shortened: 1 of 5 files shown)


exercises/csharp-2-a/.meta/test_messages.json
```diff
@@ -0,0 +1,32 @@
...
+    "cmd": "Sentence.WordWithMostVowels(\"tea\")"
+  },
+  "LongestWordIsNotWordWithMostVowels": {
+    "cmd": "Sentence.WordWithMostVowels(\"the area of a circle\")"
+  },
+  "LastWordIsWordWithMostVowels": {
+    "cmd": "Sentence.WordWithMostVowels(\"extremely fresh vegetables\")"
+  },
+  "MixedCasing": {
+    "cmd": "Sentence.WordWithMostVowels(\"TEST YOUR MIGHT\")"
```
ErikSchierboom: @iHiD Can you spot the pop-culture reference here?

iHiD: Nope! :)

iHiD: Apparently mortal compact. I never had a console with it on growing up!

ErikSchierboom: You didn't? Oh wow. It is indeed Mortal Kombat. I played that game so much 👊

iHiD: I wish I had.

README.md
```diff
@@ -18,6 +18,46 @@ These are examples of valid exercise names:
 - `javascript-2-b` : second part of exercise B for JavaScript.
 - `ruby-1-b` : first part of exercise B for Ruby.
 
+### Test messages
+
+Solving exercises that are part of the experiment is different from the regular website's process in two ways:
+
+1. Exercises are solved exclusively using in-browser coding.
+1. The test suite used to test the code is not available to the user.
+
```
iHiD: How about adding something here saying something like:

----

This means that when the code does _not_ pass one of the tests, the user won't see the code of the test that failed, or know with what input the code was called. To fix this, we need to provide a bit of context when we show errors to the user. 

For an assert_equal failure, the error with context might look as follows:

> We tried running `MyTests.MyTestMethod(1)` but got the following error:
> ```
> Expected: "foobar"
>   Actual: "barfood"
> ```

README.md
```diff
@@ -18,6 +18,46 @@ These are examples of valid exercise names:
...
+This means that when the code does _not_ pass one of the tests, the user won't know with what input the code was called. To help with this, each exercise should define a `.meta/test_messages.json` file.
```
iHiD: Add something like above then change this to?

```suggestion
To achieve with this, each exercise should define a `.meta/test_messages.json` file.
```

and then merge that with the paragraph below?

README.md
```diff
@@ -18,6 +18,46 @@ These are examples of valid exercise names:
...
+
+```json
+{
+  "MyTestMethod": {
+    "msg": "We ran `MyTests.MyTestMethod(1)` and got the following error:\n\n%{output}"
+  }
+}
+```
+
+Note that you can embed the [test runner's][test-runner-output-format] `"message"` value (usually the assertion error message) in the message by using `%{output}`.
```
iHiD:
```suggestion
Note that you can embed the [test runner's][test-runner-output-format] output (usually the assertion error message) in the message by using `%{output}`.
```

README.md
```diff
@@ -18,6 +18,55 @@ These are examples of valid exercise names:
...
 - `ruby-1-b` : first part of exercise B for Ruby.
 
+### Test messages
+
+Solving exercises that are part of the experiment is different from the regular website's process in two ways:
+
+1. Exercises are solved exclusively using in-browser coding.
+1. The test suite used to test the code is not available to the user.
+
+This means that when the code does _not_ pass one of the tests, the user won't see the code of the test that failed, or know with what input the code was called. To fix this, we need to provide a bit of context when we show errors to the user.
```
iHiD:
```suggestion
This means that when the code does _not_ pass one of the tests, the user won't see the code of the test that failed or the input the code was called. To fix this, we need to provide a bit of context when we show errors to the user.
```

PR #20, opened 2019-12-07
Add javascript/1/a

SleeplessByte:
## Instructions

You're a magician and you handle a deck of cards. In order to correctly execute your magic trick, you need to be able to _move_ a card from one position to another position. That is, you need to be able to `rearrange` the deck. Naturally, you want to be able to move in _both directions_, as well as use a negative numbers for the `from` and `to` values (so you can say "from the top" or "from the bottom" -- it's all about the magic ✨).

Create a function `arrange` that takes:

- an `array` (of length `n`) of items, 
- a `from` position (whole `number`, `-n <= from < n`), 
- and `to` position (whole `number`, `-n <= to < n`) 

It returns a _new array_ with the `item` moved from the position `from` to the position `to`:

```javascript
const before = ['🍒', '🍋', '🍎', '🍊', '🍐']
const magics = arrange(original, 1, -2)
magics
// => ['🍒', '🍎', '🍊', '🍋','🍐']
//                          ^--- has moved from position 1 to -2 (from the right side)
```
Create a second function `rearrange` that does the same thing, but **mutates** the original input `array`:

```javascript
const before = ['❤ A', '❤ 3', '❤ 6', '❤ 9', '♣ A']
const magics = rearrange(before, 4, 0)
magics
// => ['♣ A', '❤ A', '❤ 3', '❤ 6', '❤ 9']
//      ^--- has moved from position 4 to 0
magics === before
// => true
```


exercises/javascript-1-a/README.md
```diff
@@ -0,0 +1,34 @@
+## Introduction
+
+This is Part 1 of our research into how people write JavaScript differently. This exercise should take 10-15 minutes. There is no right or wrong way to approach it - just do what feels the most natural to you. Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.
+
+## Instructions
+
+You're a magician and you handle a deck of cards. In order to correctly execute your magic trick, you need to be able to _move_ a card from one position to another position. That is, you need to be able to `rearrange` the deck. Naturally, you want to be able to move in _both directions_, as well as use a negative numbers for the `from` and `to` values (so you can say "from the top" or "from the bottom" -- it's all about the magic ✨).
```
iHiD: I think move the negative numbers bits out of here, as they detract from the story/metaphor, and add them into the spec below.

```suggestion
You're a magician and you handle a deck of cards. In order to correctly execute your magic trick, you need to be able to _move_ a card from one position to another position. That is, you need to be able to `rearrange` the deck. Naturally, you want to be able to move cards in _both directions_, and be able to say "from the top of the deck" or "from the bottom of the deck".
```

exercises/javascript-1-a/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Create a function `arrange` that takes:
+
+- an `array` (of length `n`) of items,
+- a `from` position (whole `number`, `-n <= from < n`),
+- and `to` position (whole `number`, `-n <= to < n`)
+
+It returns a _new array_ with the `item` moved from the position `from` to the position `to`:
```
iHiD:
```suggestion
Positive numbers means that you move cards from the bottom of the deck (the left hand side) and negative numbers refer to starting at the top of the deck (the right hand side).

It returns a _new array_ with the `item` moved from the position `from` to the position `to`:
```

iHiD: Also, I have no idea if the top of the deck should be the LHS or the RHS. I can see strong arguments for both ways, so I think we just need to define it.

exercises/javascript-1-a/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```javascript
+const before = ['🍒', '🍋', '🍎', '🍊', '🍐']
```
iHiD: Love the use of fruits, but I do wonder if we should stick to the "card" symbols in the example below?

exercises/javascript-1-a/javascript-1-a.spec.js
```diff
@@ -0,0 +1,79 @@
+// @ts-check
+
+import { rearrange, arrange } from './javascript-1-a';
+
+const CARDS = ['\u2660', '\u2663', '\u2665', '\u2666', '\u2728']
+
+/**
+ * SLice the fruit
```
iHiD:
```suggestion
 * Slice the fruit
```

exercises/javascript-1-a/README.md
SleeplessByte: Sure!

exercises/javascript-1-a/README.md
SleeplessByte: Right. That sorta makes sense. Let's do this!

exercises/javascript-1-a/README.md
SleeplessByte: Ah yes. Everything was fruits first but then realised that windows terminals that are win32 based can't display them so switched to the cards :p

exercises/javascript-1-a/javascript-1-a.spec.js
SleeplessByte:
```suggestion
 * Prepare indices (1-based) as cards
```

exercises/javascript-1-a/README.md
```diff
@@ -0,0 +1,34 @@
...
+const magics = arrange(original, 1, -2)
+magics
+// => ['🍒', '🍎', '🍊', '🍋','🍐']
```
SleeplessByte:
```suggestion
// => ['❤ A', '❤ 3', '❤ 6', '❤ 9', '♣ A']
```

exercises/javascript-1-a/README.md
SleeplessByte:
```suggestion
const before = ['❤ A', '❤ 9', '❤ 3', '❤ 6', '♣ A']
```

SleeplessByte:
cc @ErikSchierboom

exercises/javascript-1-a/javascript-1-a.js
```diff
@@ -0,0 +1,33 @@
+// @ts-check
+
+/**
+ * Re-arranges the input array by moving an item from from, to to.
```
ErikSchierboom: I found the "from from" and "to to" bit a bit weird to read. Maybe something like this?

```suggestion
 * Re-arranges the input array by moving an item from one position to another position.
```

exercises/javascript-1-a/javascript-1-a.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {number} to the position to move to
+ *
+ * @returns {T[]}
+ */
+export function arrange(array, from, to) {
+  throw new Error('Implement the arrange function')
+}
+
+/**
+ * Re-arranges the input array by moving an item from from, to to.
```
ErikSchierboom: I found the "from from" and "to to" bit a bit weird to read. Maybe something like this?

```suggestion
 * Re-arranges the input array by moving an item from one position to another position.
```
```suggestion
 * Re-arranges the input array by moving an item from one position to another position.
```

iHiD:
@SleeplessByte Rebase and I'll merge and we can refine over the next few days if needed! :)

SleeplessByte:
@iHiD can you squash merge so I don't need to rebase :) ?

iHiD:
Apparently so!

iHiD:
@SleeplessByte Could you follow up with an equivalent of https://github.com/exercism/research_experiment_1/pull/24 pls?

PR #21, opened 2019-12-07
x86-64-assembly/a: add part 1

bergjohan:
## Instructions

Your task is to determine if all characters in a string are alphabetic, i.e., either an uppercase letter, or a lowercase letter.

Your job is to define a function with the signature `str_isalpha`, which takes a string, and returns a non-zero value if all characters in the string are alphabetic and there is at least one character, zero otherwise.

For example:

```c
str_isalpha("Hello");
// Returns: non-zero
```


bergjohan:
Thanks! :) Yeah, I'll wait, I found an error while reviewing my own code :P

iHiD:
If you rebase, I'll merge, and then we can just improve it over the next few days as necessary

bergjohan:
Cool! I saw that @ErikSchierboom renamed the file `test_messages.json` to `config.json`, and made a few changes to it. I'll try to rebase and make those changes later tonight.

iHiD:
Great. Thanks :)

bergjohan:
Rebased and changed to `config.json`

iHiD:
Thanks! :)

PR #22, opened 2019-12-09
Sample Ruby exercise to go with new schema

iHiD:
This is a port of the C# exercise, purely for testing. This will not be the final Ruby exercise.

This conforms to a slightly updated schema that @kntsoriano and I are working on, and is intended to go with https://github.com/exercism/website/pull/849


exercises/ruby-1-a/.meta/config.json
```diff
@@ -0,0 +1,34 @@
+{
+  "solution_files": ["ruby_1_a.rb"],
+  "test_file": "ruby_1_a_test.rb",
+  "tests": [
```
ErikSchierboom: I like the new tests entry. The only disadvantage I can think of is that people could forget to update these. So maybe some sort of warning mechanism that can detect if this section is missing a key would be nice later.

exercises/ruby-1-b/.meta/config.json
```diff
@@ -0,0 +1,35 @@
+{
+  "solution_files": ["ruby_1_b.rb"],
```
ErikSchierboom: Just to confirm. You still need to update this file, right? It is now the same as exercise A.

exercises/ruby-1-a/ruby_1_a_example.rb
```diff
@@ -1 +1,11 @@
-TODO
\ No newline at end of file
+VOWELS = %w{a e i o u}
+def word_with_most_vowels(sentence)
```
ErikSchierboom: Have you tried to see if there are multiple valid approaches?

exercises/ruby-1-a/.meta/config.json
iHiD: Yes. Agreed. I could also definitely generate this for Ruby, and/or add a CI test for Ruby for this. I don't know how generalisable that is.

I am wondering if a test generator could output both files automatically.

exercises/ruby-1-b/.meta/config.json
iHiD: Correct. Both exercises are identical. Purely there so we can all beta test on Ruby without it mattering which exercise we get assigned, basically.

exercises/ruby-1-a/ruby_1_a_example.rb
iHiD: Nope. I just blindly copied what you've done and changed it to Ruby :)

This won't be the actual Ruby exercise so I'm not worrying too much. I imagine there would be though. I would definitely not use a single method for example.

exercises/ruby-1-a/.meta/config.json
ErikSchierboom: The C# test generator could definitely do this, and I expect most could.

exercises/ruby-1-b/.meta/config.json
ErikSchierboom: Okay cool!

PR #26, opened 2019-12-11
csharp: add troubleshooting instruction

ErikSchierboom:


exercises/csharp-1-a/Csharp1A.cs
```diff
@@ -1,7 +1,9 @@
+using System;
```
ErikSchierboom: This helps people just do `Console.WriteLine` instead of `System.Console.WriteLine`.

PR #33, opened 2019-12-19
x86-64-assembly/b: add part 1

bergjohan:
## Introduction

This is Part 1 of our research into how people write x86-64 assembly differently. This exercise should take 10-15 minutes. There is no right or wrong way to approach it - just do what feels the most natural to you. Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.

## Instructions

Your job is to define a function with the signature `color_name`, which takes an enum value, and returns the name of the constant in the enumeration that has the specified value.

For example:

```c
color_name(RED);
// Returns: "RED"
```

The enum is declared as follows:

```c
enum color {
    RED,
    ORANGE,
    YELLOW,
    GREEN,
    BLUE,
    INDIGO,
    VIOLET
};
```


ErikSchierboom:
This is very nice! Do you have any idea what the second part will look like?

bergjohan:
Thanks! :)

My idea is to let the student extend the function to copy the color name into a user-provided buffer instead of returning the string, see [`strerror_s`](https://en.cppreference.com/w/c/string/byte/strerror) for an example of this.

There are a few different ways to copy the string in assembly. Although they wont be tied to a specific approach from part 1, I still hope this is good enough for the experiment.

It's been a lot harder than I first thought to come up with exercises that can be solved using multiple different approaches :)

ErikSchierboom:
Sounds good! I also found coming up with exercises that can be solved using different approaches was hard!

bergjohan:
Great! I'll try to get the PR ready in the next couple of days!

PR #36, opened 2019-12-25
x86-64-assembly/b: add part 2

bergjohan:
## Introduction

This is Part 2 of our research into how people write x86-64 assembly differently. This exercise should take 10-15 minutes. There is no right or wrong way to approach it - just do what feels the most natural to you. Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.

## Instructions

Your task is to extend existing code with new functionality. The code you'll be working with has a function named `color_name`. This function takes an enum value, and returns the name of the constant in the enumeration that has the specified value.

For example:

```c
color_name(RED);
// Returns: "RED"
```

The enum is declared as follows:

```c
enum color {
    RED,
    ORANGE,
    YELLOW,
    GREEN,
    BLUE,
    INDIGO,
    VIOLET
};
```

Your job is to extend this code to copy the name into a user-provided storage buf.

For example:

```c
char buf[BUFSIZE];

color_name(buf, RED);
// buf contains "RED"
```


exercises/x86-64-assembly-2-b/README.md
```diff
@@ -0,0 +1,39 @@
...
+    ORANGE,
+    YELLOW,
+    GREEN,
+    BLUE,
+    INDIGO,
+    VIOLET
+};
+```
+
+Your job is to extend this code to copy the name into a user-provided storage buf.
```
ErikSchierboom: Maybe add a little bit of text explaining that the existing `color_name` function needs to be modified (and also _how_ it should be modified)?

exercises/x86-64-assembly-2-b/.meta/config.json
```diff
@@ -0,0 +1,34 @@
+{
+  "solution_files": ["x86_64_assembly_2_b.asm"],
+  "test_file": "x86_64_assembly_2_b_test.c",
+  "tests": [
+    {
+      "name": "test_red",
+      "cmd": "buf"
```
ErikSchierboom: Maybe I misread, but isn't the actual command:

```suggestion
      "cmd": "color_name(buf, RED)"
```

bergjohan: Lol, I have no idea what I was thinking when I wrote this :-P

bergjohan:
I updated the PR based on your comments, please tell me what you think!

ErikSchierboom:
Thanks again @bergjohan!

Issue #35, opened 2019-12-23
x86-64 assembly function description


bergjohan:
In x86-64 assembly, function parameters are passed in specific registers. If you are new to assembly, and don't know about [calling conventions](https://en.wikipedia.org/wiki/Calling_convention), it might be difficult to get started.

Maybe we should add a comment to the stub file, describing the function and its registers, to make it a bit easier.

Example:

```nasm
;
; int str_isalpha(const char *str);
;
; Parameters:
;   rdi - str
; Returns:
;   rax - a non-zero value if all characters in the string are alphabetic and there is at least one character, zero otherwise.
;
section .text
global str_isalpha
str_isalpha:
    ; Provide your implementation here
    ret
```

What do you think?

bergjohan:
Or we could describe the registers in the README file, that might be better?

ErikSchierboom:
I think it could work both ways, although I'd prefer it in the README I think.

PR #37, opened 2019-12-25
x86-64-assembly/a: Update README for part 1

bergjohan:
Ref: #35

### DIFF
--- a/exercises/x86-64-assembly-1-a/README.md
+++ b/exercises/x86-64-assembly-1-a/README.md
@@ -6,7 +6,7 @@ This is Part 1 of our research into how people write x86-64 assembly differently
 
 Your task is to determine if all characters in a string are alphabetic, i.e., either an uppercase letter, or a lowercase letter.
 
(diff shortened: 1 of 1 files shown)


exercises/x86-64-assembly-1-a/README.md
```diff
@@ -6,11 +6,13 @@ This is Part 1 of our research into how people write x86-64 assembly differently
...
+Your job is to define a function with the signature `str_isalpha`, which takes a pointer to a null-terminated byte string, and returns a non-zero value if all characters in the string are alphabetic and there is at least one character, zero otherwise.
 
 For example:
 
 ```c
 str_isalpha("Hello");
 // Returns: non-zero
 ```
+
+The string is passed in the `rdi` register, and the return value in `rax`.
```
ErikSchierboom: Just a tiny suggestion:

```suggestion
The string is passed in the `rdi` register, and the return value should be in the `rax` register.
```

But feel free to ignore and merge if you'd like.

ErikSchierboom:
Merged! Thanks @bergjohan!

PR #38, opened 2019-12-27
x86-64-assembly/a: Update README for part 2

bergjohan:
Ref: #35


bergjohan:
@iHiD Updated this PR with the suggested changes from #37

PR #39, opened 2019-12-27
x86-64-assembly/b: Update README for part 1

bergjohan:
Ref: #35

### DIFF
--- a/exercises/x86-64-assembly-1-b/README.md
+++ b/exercises/x86-64-assembly-1-b/README.md
@@ -26,3 +26,5 @@ enum color {
     VIOLET
 };
 ```
+
+The enum value is passed into the function via the `rdi` register, and the return value should be stored in the `rax` register.


exercises/x86-64-assembly-1-b/README.md
```diff
@@ -26,3 +26,5 @@ enum color {
     VIOLET
 };
 ```
+
+The enum value is passed in the `rdi` register, and the return value should be in the `rax` register.
```
iHiD:
```suggestion
The enum value is passed into the `rdi` register, and the return value should be stored in the `rax` register.
```

This reads slightly better I think but it might be incorrect.

bergjohan: Thank you for the review! :)

What I'm trying to explain is the following:

The caller will put the first function argument in the `rdi` register before calling the function.
The callee will know to look for the first function parameter in the `rdi` register.

Kind of like if you want to mail a letter, you put the letter in an envelope before sending the mail.
The receiver opens the envelope to read the letter.

Saying that the letter is passed into the envelope doesn't sound right to me.

Maybe something like "The `rdi` register is used to pass the enum value to the function" would be better?

I like the suggestion to add the word "stored" :)

iHiD: > Maybe something like "The rdi register is used to pass the enum value to the function" would be better?

Yeah - I like that. Or maybe "it is passed into the function via the `rdi` register".

bergjohan:
@iHiD I've updated the PR with your suggested changes.

ErikSchierboom:
@iHiD I've taken the liberty to merge this, as your suggestions were implemented :)

PR #49, opened 2020-01-13
Add first part of Elm exercise b

mpizenberg:
While the exercise A was a data manipulation problem (lists, dicts, etc.) this one is a parsing problem, aiming at turning a string into a list of structured data (a `List Pizza`). It should take 10 to 15 minutes to an Elm dev.

I have ideas on how to increase the difficulty for the second part, for example adding ingredients of pizzas, which are in an unknown number for each pizza when calling the parsing function.

### DIFF
--- a/exercises/elm-1-b/README.md
+++ b/exercises/elm-1-b/README.md
@@ -1,0 +1,35 @@
+## Introduction
+
+This is Part 1 of our research into how people write Elm differently. This exercise should take 10-15 minutes. There is no right or wrong way to approach it - just do what feels the most natural to you. Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.
+
+## Instructions
+
+You are the proud owner of an amazing pizzeria but your website is missing the menu.
+So you scanned a printed menu and the scanner was able to extract the text in the image.
+Unfortunately it is all one big block of text like follows:
+
+```elm
+scannedMenu = "formaggio (v) 8€, regina 11€, ..."
+```
+
+At least it follows a useful patern.
+There is the name, an indication if it is vegetarian, and the price.
+**Your task is to convert this string into a list of pizzas.**
+The `Pizza` type is already provided for convenience.
+
+## Elm Documentation
+
+In order to save time googling, some links to the Elm documentation are here. This is designed to be a fairly exhaustive list, as opposed to a list of things that any one solution will need
+
+- [Debug](https://package.elm-lang.org/packages/elm/core/latest/Debug)
+- [List](https://package.elm-lang.org/packages/elm/core/latest/List)
+- [Dict](https://package.elm-lang.org/packages/elm/core/latest/Dict)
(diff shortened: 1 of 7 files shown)


exercises/elm-1-b/src/Pizza.elm
```diff
@@ -0,0 +1,13 @@
...
+type alias Pizza =
+    { name : String
+    , vegetarian : Bool
+    , price : Int
+    }
+
+
+retrieveMenu : String -> List Pizza
+retrieveMenu stringMenu =
+    Debug.todo "retrieveMenu"
```
ceddlyburge: Hi Matthieu, this is meant to return "Please implement the `retrieveMenu` function". It is good if this makes it in to the output somehow, although it looks like this will be difficult in this case.

Cheers, Cedd

exercises/elm-1-b/README.md
```diff
@@ -0,0 +1,35 @@
+## Introduction
+
+This is Part 1 of our research into how people write Elm differently. This exercise should take 10-15 minutes. There is no right or wrong way to approach it - just do what feels the most natural to you. Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.
+
+## Instructions
+
+You are the proud owner of an amazing pizzeria but your website is missing the menu.
```
ErikSchierboom: Love this background story!

exercises/elm-1-b/src/Pizza.elm
mpizenberg: Ok, I've updated the `Debug.todo` message to say exactly that. If I've understood correctly (@ErikSchierboom?), exercism will display the error in the message returned to the student. In this case the error becomes:
> This test failed because it threw an exception: "Error: TODO in module `Pizza` on line 13
> Please implement the `retrieveMenu` function"

ErikSchierboom: @mpizenenberg Well, if the test runner correctly captures the error and returns it as part of the `results.json` document, then yes.

exercises/elm-1-b/README.md
ceddlyburge: Hi Erik, I've already taken a look at this, and have just approved it. Not that I think matthieu really needs me to check his work :)

exercises/elm-1-b/src/Pizza.elm
ceddlyburge: Charles care is integrating the test runner. I emailed him last week to see if there was anything we could do to help, but I don't think there is (not yet anyway). He was waiting for a website update in order to do something. It seems to work locally for me, but obviously I can't test it within the exercism system.

ErikSchierboom:
Merged! Thanks a lot @mpizenberg and @ceddlyburge!

exercises/elm-1-b/README.md
mpizenberg: we always need another pair of eyes ;)

PR #51, opened 2020-01-14
Add Elm and Go exercises

iHiD:
@ErikSchierboom - We need to remember to do this. Could you check the file is a correct representation of the current state tomorrow pls.

### DIFF
--- a/config.json
+++ b/config.json
@@ -40,6 +40,38 @@
       "topics": []
     },
     {
+      "slug": "elm-1-a",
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@iHiD Good catch. Must have been confused by exercises now also having a config.json file. I'll check the current state.

ErikSchierboom:
@iHiD Fixed in #52

PR #57, opened 2020-01-20
Add add ruby-1-a

pvcarrera:
This PR contains a proposal for the first leg of the Ruby experiment `ruby_1_a`. For the second leg, the idea is to play around with discounts but it's still not fully defined.

Please let me know wha your thoughts are.


exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,18 @@ This is Part 1 of our research into how people write Ruby differently. This exer
 
 ## Instructions
 
-Your task is to determine which word in a sentence has the most vowels (a, e, i, o, u). A sentence is defined as a [Ruby string][docs-string] consisting of one or more words. A word consists of one or more upper- or lowercase letters. Words are separated from each other by a single space character.
-
-Your job is to define a method with the signature `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+Your task is to define a ShoppingCart that can receive a hash of items and has two methods: `total_amount` and `items_list`. The method `total_amount` will return the sum of all the items’ prices. The method `list` will return a string with the name of the items separated by commas.
```
kotp: Describe the "hash of items" to include the key and value expected.  I am not sure the user will see the test material to get that hint.  Hopefully we aren't taunting them with less than the needed knowledge required up front.

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,18 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
 
 For example:
 
 ```ruby
-word_with_most_vowels("Hi there aunt jeanie")
-// Returns: "jeanie"
-```
+shopping_cart = ShoppingCart.new({ potatoes: 10.00, bananas: 30.00 })
```
kotp: This covers that comment made with a code comment, but we should repeat it in code, making the initial declaration in words as we do "items".

exercises/ruby-1-a/ruby_1_a.rb
```diff
@@ -1,3 +1,13 @@
-def word_with_most_vowels(sentence)
-  # Start your code here.
+class ShoppingCart
+  def initialize(items)
+    # Start your code here.
```
kotp: We don't really need the "start" word here, I think.

exercises/ruby-1-a/ruby_1_a_example.rb
```diff
@@ -1,11 +1,17 @@
...
-      word.downcase.chars.select{|l|VOWELS.include?(l)}.count
-    ]
+class ShoppingCart
+  def initialize(items = {})
+    @items = items
   end
-  word_vowels.sort_by {|w,c| -c}.first.first
+
+  def total_amount
+    items.sum(0) { |key, value| value }
```
kotp: item, cost`
```suggestion
    items.sum(0) { |_item, cost| cost }
```

It is idiomatic to prefix unused variables with an underscore, or as an underscore completely.

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,18 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
-[docs-string]: https://ruby-doc.org/core-2.4.1/String.html
+shopping_cart.total_amount
+// Returns: 40
+
+shopping_cart.items_list
+// Returns: "potatoes, bananas"
+```
 
-**Note:** You can use `debug("some test")` to output debugging information for yourself.
+**Note:** You can use debug("some test") to output debugging information for yourself.
```
ErikSchierboom: Any particular reason not to use backticks around the debug code?

exercises/ruby-1-a/ruby_1_a.rb
ErikSchierboom: I don't know if this is possible in Ruby, but ideally we'd put some code here that would provide a nice error message when the student submits the code _as is_.

Here are two examples of how the C# and Go experiment solve this:

1. C#: https://github.com/exercism/research_experiment_1/blob/e60a36db7510030cb34e4b5e32e17a4baed802d5/exercises/csharp-1-a/Csharp1A.cs#L7
1. Go: https://github.com/exercism/research_experiment_1/blob/e60a36db7510030cb34e4b5e32e17a4baed802d5/exercises/go-1-a/go_1_a.go#L5

I don't know if  something like this is possible in Ruby?

exercises/ruby-1-a/README.md
ErikSchierboom: Maybe:

```suggestion
Your task is to define a `ShoppingCart` class that can receive a hash of items and has two methods: `total_amount` and `items_list`. The method `total_amount` will return the sum of all the items’ prices. The method `list` will return a string with the name of the items separated by commas.
```

exercises/ruby-1-a/ruby_1_a_example.rb
```diff
@@ -1,11 +1,17 @@
-VOWELS = %w{a e i o u}
-def word_with_most_vowels(sentence)
-  words = sentence.split(" ")
-  word_vowels = words.map do |word|
-    [
-      word,
-      word.downcase.chars.select{|l|VOWELS.include?(l)}.count
-    ]
+class ShoppingCart
```
ErikSchierboom: For the second part of the experiment, there will be three starting points, each using different approaches. Could you perhaps add two new example implementations, using different approaches? That way, you immediately have the starting points for the second part of the exercise, and a verification that it can indeed be solved using different approaches.

exercises/ruby-1-a/README.md
pvcarrera: No, I'm not sure why I removed them, probably copy & paste from the wrong place

pvcarrera:
Thanks for your reviews @kotp @ErikSchierboom, I'll try to address your comments during the day.

@ErikSchierboom I'm not sure why bu I can't reply to your error messages comment, but I like your idea. In Ruby we can: `fail NotImplementedError.new('Please implement etc.')`

exercises/ruby-1-a/ruby_1_a_example.rb
pvcarrera: So the plan is to have three examples for the same test suite: `ruby_1_a_example_1.rb, ruby_1_a_example_2.rb, ruby_1_a_example_3.rb`, right?

ErikSchierboom: Correct!

exercises/ruby-1-a/ruby_1_a.rb
```diff
@@ -1,3 +1,13 @@
-def word_with_most_vowels(sentence)
-  # Start your code here.
+class ShoppingCart
+  def initialize(items)
+    fail NotImplementedError.new('Please implement the ShoppingCart initializer')
```
ErikSchierboom: I'm not that well versed in Ruby, but is "initializer" something people will be familiar with? Instead of doing:

```suggestion
    fail NotImplementedError.new('Please implement the ShoppingCart#initialize method')
```

ErikSchierboom: I like the `NotImplementedError`!

pvcarrera: In Ruby when you call `Class.new`, Ruby calls the `initialize` method. The method new will create a new instance of the class, and call initialize on it.

That's the reason why I wasn't sure about `ShoppingCart#initialize` or `ShoppingCart#new` but I guess is fine to go with `ShoppingCart#initialize`

ErikSchierboom: Thanks for explaining.

exercises/ruby-1-a/ruby_1_a_example.rb
pvcarrera: Ok, added another two examples keeping the same interface and tests.

exercises/ruby-1-a/ruby_1_a_test.rb
```diff
@@ -1,33 +1,20 @@
 require 'minitest/autorun'
-require_relative 'ruby_1_a'
+require_relative 'ruby_1_a_example_3'
```
ErikSchierboom: This should be reverted (probably a remnant of you testing the third example :))

```suggestion
require_relative 'ruby_1_a'
```

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,18 @@ This is Part 1 of our research into how people write Ruby differently. This exer
 
 ## Instructions
 
-Your task is to determine which word in a sentence has the most vowels (a, e, i, o, u). A sentence is defined as a [Ruby string][docs-string] consisting of one or more words. A word consists of one or more upper- or lowercase letters. Words are separated from each other by a single space character.
-
-Your job is to define a method with the signature `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+Your task is to define a `ShoppingCart` class that can receive a hash of items where the key is the name of the item and the value the price of the item, see the example below. The `ShoppingCart` class will also have two methods: `total_amount` and `items_list`. The method `total_amount` will return the sum of all the items’ prices. The method `list` will return a string with the name of the items separated by commas.
```
iHiD: Could we add a little overview before digging into the technical spec?

Something like:

> Your task is to develop a shopping cart, which can contain items and calculate total cart price. This will be modelled as `ShoppingCart` class...

iHiD: >  The method `list` 

Should this be `items_list`?

pvcarrera: Thanks for your feedback. I added a small overview and fixed the method name

iHiD:
@pvcarrera Thanks for doing this! I'm really excited to have Ruby in the Experment :)

I've given this some thought. I really like the shopping cart vibe as an exercise, but I'm a bit concerned that there are currently not a wide variety of ways that it could be solved, and that therefore as a basis for our research it will not give us a good variety of data. 

I've added a PR to make it a bit more complex, and try and generate some more possible variants: https://github.com/pvcarrera/research_experiment_1/pull/1

I've PR'd that to your branch, so we can decide if that approach is good or not before updating what you have here. I describe a bit more about the approaches in the commit message there. Keen to hear your thoughts!

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,34 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
 
-Your job is to define a method with the signature `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+Items are added to the basket using stock-keeping unit (SKU) codes. The items in the shop are:
+
+| Name      | Price | SKU    |
+|-----------|-------|--------|
+| Potatoes  | 10    | D69856 |
+| Rice      | 30    | F55690 |
+| Cofee     | 14.99 | C662F6 |
+| Newspaper | 2.99  | B48C0D |
```
kotp: If possible define the SKU with a prefix code that might make sense.  If I want all staples, or all media, perhaps I can gather all available products that start with S or M.  It mirrors internal sku usage (at least for our jewelry company) so that we can compare items to the customer as well "You viewed this, you may be interested in these!" type of an ability.  At the moment, at least to me, the SKU doesn't seem to make sense.

pvcarrera: I guess at the moment it's only an identifier something like an UUID. I think we can add the prefix it's not going to do any harm and it can open some possibilities for extensions.

pvcarrera: I updated the SKUs so now they follow the convention `[3 chars from product type][3 chars from the product name]`

iHiD: It was `hex(3).upcase` :)

exercises/ruby-1-a/ruby_1_a.rb
```diff
@@ -1,3 +1,19 @@
-def word_with_most_vowels(sentence)
-  # Start your code here.
+# frozen_string_literal: true
+
+class ShoppingCart
+  def initialize
```
iHiD: I think we can delete the `initialize` method from this file.

iHiD:
@ErikSchierboom @kotp Could you both give this a final look over pls. Then let's get it merged if you're both happy?

@pvcarrera @kotp Did you have an idea for a second exercise too, which we could work on?

exercises/ruby-1-a/ruby_1_a_example_1.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class ShoppingCart
+  STOCK = {
+    STAPOT: ['Potatoes', 10.0],
+    STARIC: ['Rice', 30.00],
+    STACOF: ['Coffee', 14.99],
+    MEDNEW: ['Newspaper', 2.99]
```
kotp: The SKU numbers used here should be reflected in the instructions.  No value in not giving this as factual data.

pvcarrera: Yep, now the README also reflects these changes

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,34 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
+
+You need to define a `ShoppingCart` class, with three public methods:
+- `add`: Takes an SKU, and adds it to the cart
```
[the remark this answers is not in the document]
pvcarrera: 👍 I added the punctuation fixes to the last commit

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,34 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
+- `items_list`: A list of the names of the items in the cart, separated by commas.
 
 For example:
 
 ```ruby
-word_with_most_vowels("Hi there aunt jeanie")
-// Returns: "jeanie"
-```
+shopping_cart = ShoppingCart.new
+shopping_cart.add(D69856)
```
ErikSchierboom: I think this needs to be updated too.

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,34 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
+shopping_cart.add(F55690)
```
ErikSchierboom: I think this needs to be updated too.

exercises/ruby-1-a/.meta/config.json
```diff
@@ -3,32 +3,32 @@
...
-      "name": "test_one_word_with_one_vowel",
-      "cmd": "word_with_most_vowels(\"at\")"
+      "name": "test_one_item",
+      "cmd": "ShoppingCart.new.total_amount"
     },
     {
-      "name": "test_one_word_with_one_vowel_and_one_consonant",
-      "cmd": "word_with_most_vowels(\"at\")"
+      "name": "test_initial_total_amount",
+      "cmd": "ShoppingCart.new.add(:D69856).total_amount"
```
ErikSchierboom: Needs to be updated too

exercises/ruby-1-a/.meta/config.json
```diff
@@ -3,32 +3,32 @@
...
     },
     {
-      "name": "test_one_word_with_two_vowels_and_one_consonant",
-      "cmd": "word_with_most_vowels(\"tea\")"
+      "name": "test_two_items",
+      "cmd": "ShoppingCart.new.add(:D69856).add(:F55690).total_amount"
```
ErikSchierboom: Needs to be updated too

exercises/ruby-1-a/.meta/config.json
```diff
@@ -3,32 +3,32 @@
...
     },
     {
-      "name": "test_longest_word_is_not_word_with_most_vowels",
-      "cmd": "word_with_most_vowels(\"the area of a circle\")"
+      "name": "test_list_items",
+      "cmd": "ShoppingCart.new.add(:D69856).add(:F55690).list_items"
```
ErikSchierboom: Needs to be updated too

exercises/ruby-1-a/.meta/config.json
```diff
@@ -3,32 +3,32 @@
...
     },
     {
-      "name": "test_last_word_is_word_with_most_vowels",
-      "cmd": "word_with_most_vowels(\"extremely fresh vegetables\")"
+      "name": "test_items_ordered_alphabetically",
+      "cmd": "ShoppingCart.new.add(:F55690).add(:D69856).list_items"
```
ErikSchierboom: Needs to be updated too

exercises/ruby-1-a/.meta/config.json
```diff
@@ -3,32 +3,32 @@
...
     },
     {
-      "name": "test_mixed_casing",
-      "cmd": "word_with_most_vowels(\"TEST YOUR MIGHT\")"
+      "name": "test_duplicate_items_sums_price",
+      "cmd": "ShoppingCart.new.add(:F55690).add(:F55690).list_items"
```
ErikSchierboom: Needs to be updated too

exercises/ruby-1-a/.meta/config.json
```diff
@@ -3,32 +3,32 @@
...
     },
     {
-      "name": "test_ignore_word_without_vowels",
-      "cmd": "word_with_most_vowels(\"great rhythm\")"
+      "name": "test_duplicate_items_uniqs_list",
+      "cmd": "ShoppingCart.new.add(:F55690).add(:F55690).list_items"
```
ErikSchierboom: Needs to be updated too

pvcarrera: Sorry I need to get better at search and replace 😊

exercises/ruby-1-a/ruby_1_a_test.rb
```diff
@@ -1,33 +1,53 @@
...
+    cart.add(:STARIC)
+    cart.add(:STAPOT)
+    assert_equal 'Potatoes, Rice', cart.items_list
   end
 
-  def test_last_word_is_word_with_most_vowels
-    assert_equal "vegetables", word_with_most_vowels("extremely fresh vegetables")
+  def test_items_ordered_alphabetically
+    cart = ShoppingCart.new
+    cart.add(:STAPOT)
```
iHiD: I think these need to be swapped round, so that the order added to the basket is **not** alphabetical. So Rice should be added before Potatoes. (And we should switch it in the previous test also)

exercises/ruby-1-a/README.md
```diff
@@ -4,17 +4,34 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
+|-----------|-------|--------|
+| Potatoes  | 10    | STAPOT |
+| Rice      | 30    | STARIC |
+| Coffee    | 14.99 | STACOF |
+| Newspaper | 2.99  | MEDNEW |
+
+You need to define a `ShoppingCart` class, with three public methods:
+- `add`: Takes an SKU and adds it to the cart. The SKU is comprised of the first three upcased chars of the product type + the first three upcased chars of the product name.
+- `total_amount`: The value of the items in the cart as a number with two decimal places (e.g. 3.99).
+- `items_list`: A list of the names of the items in the cart, separated by commas.
```
iHiD:
```suggestion
- `items_list`: A list of the names of the items in the cart, ordered alphabetically, separated by a comma and a space.
```

iHiD:
Two small comments, then I think we're there. Really happy with how this has turned out.

pvcarrera:
Yep, thanks to all of you this exercise is looking really nice

ErikSchierboom:
Merged! Thanks a ton @pvcarrera.

PR #58, opened 2020-01-24
Add ruby-1-b

goalaleo:
This is the first draft of the first part of ruby-1-b exercise.
Exercise description is in the `README.md`. Feedback much appreciated :) !

The idea for the second part is to implement a search based on books' fields

Co-author: [@kotp](https://github.com/kotp)


exercises/ruby-1-b/README.md
```diff
@@ -1,20 +1,27 @@
...
+to use books, Google or Stack Overflow, just like you would if you were
+programming normally, but please don't actively look for a solution to
+the problem as this will negatively affect the research.
 
 ## Instructions
 
-Your task is to determine which word in a sentence has the most vowels (a, e, i, o, u). A sentence is defined as a [Ruby string][docs-string] consisting of one or more words. A word consists of one or more upper- or lowercase letters. Words are separated from each other by a single space character.
+You are given a collection of library books, and your job is to import them into
+the library's catalog system. The books are in a text file which provides a title,
+ISBN, author, and publishing date for each book.
```
goalaleo:
```suggestion
ISBN, author, and publishing year for each book.
```

exercises/ruby-1-b/ruby_1_b_example.rb
```diff
@@ -1,12 +1,23 @@
...
-    ]
+class Library
+  def self.books
+    @books ||= import_books
+  end
+
+  def self.import_books
+    file_name = "books.txt"
+    IO.readlines(file_name, chomp: true).drop(1)
+      .map { |line| line.split(";") }
```
pvcarrera: I prefer single quoted strings when there is no need for string interpolation or special symbols but I know this is a style preference so it's up to you. Do you know if we use Rubocop or something similar?

exercises/ruby-1-b/ruby_1_b_test.rb
```diff
@@ -1,34 +1,32 @@
...
-    assert_equal "a", word_with_most_vowels("a")
+class LibraryTest < Minitest::Test
+  def test_first_book_imported
+    assert Library.books.map(&:isbn).include?("9780191604928")
   end
 
-  def test_one_word_with_one_vowel_and_one_consonant
-    assert_equal "at", word_with_most_vowels("at")
+  def test_last_book_imported
+    assert Library.books.map(&:isbn).include?("9780451524935")
```
pvcarrera: Assert `assert_equal` provides better messages. Wen this test fails we get an error that says `Expected false to be truthy.`. With something like:
```
assert_equal '9780191604928', Library.books.first.isbn
```
We'll get: 
```
Expected: "9780191604928"
  Actual: "i9780191604928"
```

exercises/ruby-1-b/ruby_1_b_example.rb
goalaleo: I personally try to use double quotes always (although sometimes some single quotes slip in there). But I'm not very pedantic about this issue, so if there's some project style guide that prefers the single quote style, then I'm fine with changing this

goalaleo:
Thanks @pvcarrera ! I made some changes based on your feedback

exercises/ruby-1-b/ruby_1_b_example.rb
kotp: As long as they are [consistent](https://rubystyle.guide/#consistent-string-literals) I don't mind either, though I prefer single quotes by default.

exercises/ruby-1-b/README.md
```diff
@@ -1,20 +1,27 @@
...
 
-Your job is to define a method with the signature `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+Write the methods that will import the data for the books from the text file to
+the library's catalog system.
 
-For example:
+The data is stored in plain text in a file called `books.txt`. One record for each line,
+and the `;` character is used as a delimiter. The first two lines look like this:
 
-```ruby
```
ErikSchierboom: Could you perhaps also add some example code?

exercises/ruby-1-b/ruby_1_b_example.rb
```diff
@@ -1,12 +1,24 @@
-VOWELS = %w{a e i o u}
-def word_with_most_vowels(sentence)
-  words = sentence.split(" ")
-  word_vowels = words.map do |word|
-    [
-      word,
-      word.downcase.chars.select{|l|VOWELS.include?(l)}.count
-    ]
+class Library
```
ErikSchierboom: My slight worry is that this exercise might be a little bit too simple. I'm not a Ruby programmer though. Maybe @iHiD has some thoughts?

exercises/ruby-1-b/README.md
```diff
@@ -1,20 +1,27 @@
...
 
 ## Instructions
 
-Your task is to determine which word in a sentence has the most vowels (a, e, i, o, u). A sentence is defined as a [Ruby string][docs-string] consisting of one or more words. A word consists of one or more upper- or lowercase letters. Words are separated from each other by a single space character.
+You are given a collection of library books, and your job is to import them into
+the library's catalog system. The books are in a text file which provides a title,
+ISBN, author, and publishing year for each book.
 
-Your job is to define a method with the signature `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+Write the methods that will import the data for the books from the text file to
```
ErikSchierboom: I'm missing the bit that explains what code the student is expected to write.

exercises/ruby-1-b/books.txt
```diff
@@ -0,0 +1,10 @@
+Title;ISBN;Author;Published
+The Adventures of Tom Sawyer;9780191604928;Mark Twain;2007
+Republic;9780718198916;Plato;2012
+Programming Ruby: The Pragmatic Programmers' Guide;9780974514055;David Thomas;2004
+Pride and Prejudice by Jane Austen;9781986431484;Jane Austen;2018
+To Kill a Mockingbird;9780446310789;Harper Lee;1988
+Cosmicomics;9780330319089;Italo Calvino;1969
+The Lord of the Rings;9780544003415;J. R. R. Tolkien;2012
+Lord of the Flies;9780140283334;William Golding;1999
+1984: A Novel;9780451524935;George Orwell;2009
```
kytrinyx: We should probably avoid using external files, since they would not be able to see what's in the file.

How about providing the file as a string inside the stub of the solution file?

kytrinyx:
I like the library idea. I'm a fan of books, so I guess I'm biased :-)

Could we perhaps add a little bit of complexity so that importing the string into some sort of actual structure becomes an implicit design choice that the person has to make?

Then maybe add some complexity like:

- a library can have multiple copies of a book
- you can borrow books (but only if there is at least one copy that is not lent out)
- you can return books
- you can ask for recommendations that are newer than a certain year (the result would be title + author, ordered alphabetically by something or other)
- you can get a count of how many books have been lent out in total at any one time

That would give us a bit more to work with in this iteration of the research project, as well as a bunch of ways to expand on it for the second phase (adding arbitrary constraints, etc).

exercises/ruby-1-b/books.txt
kotp: Why would they not be able to see the contents if they are able to read it?  Doesn't `debug` give us that ability?

kotp: I am testing `debug` as described in the first one available (shopping cart) and it doesn't seem to work yet... not seeing anything when I use it and run the tests.

ErikSchierboom:
@goalaleo Have you perhaps had time to look at the review comments?

ErikSchierboom:
@goalaleo Bump :)

goalaleo:
Sorry about the delay @ErikSchierboom I modified the exercise based on the feedback. I think that it's at least complex enough now for 10-15 minutes of work.

I'm not sure how detailed the instructions should be - is it enough to give the students just enough information to get started on the exercise, and then let the test suite output instruct them on what to do next?

exercises/ruby-1-b/.meta/config.json
```diff
@@ -3,39 +3,72 @@
   "test_file": "ruby_1_b_test.rb",
   "tests": [
     {
-      "test": "test_one_word_with_one_vowel",
-      "name": "one word with one vowel",
-      "cmd": "word_with_most_vowels(\"at\")"
+      "name": "test_find_by_empty_isbn_returns_nil",
+      "cmd": "@library.find(\"\")"
```
ErikSchierboom: In the `ruby-1-a` example, we're giving the full code used to test the input. See https://github.com/exercism/research_experiment_1/blob/master/exercises/ruby-1-a/.meta/config.json#L14.

Apply that to this test, this would become:

```suggestion
      "cmd": "library = Library.new\nlibrary.find(\"\")"
```

Could you update the other entries too?

exercises/ruby-1-b/.meta/config.json
```diff
@@ -3,39 +3,72 @@
   "test_file": "ruby_1_b_test.rb",
   "tests": [
     {
-      "test": "test_one_word_with_one_vowel",
```
ErikSchierboom: The `"test"` property should contain the name as it is found in the test runner, which is usually the test name. In this case, `"test"` should thus equal `"test_find_by_empty_isbn_returns_nil"`. The `"name"` property is required and should contain a human-readable version of the test. In this case, probably something like `"find by empty isbn returns nil"`.

Could you update the other entries too?

exercises/ruby-1-b/.meta/config.json
```diff
@@ -3,39 +3,72 @@
   "test_file": "ruby_1_b_test.rb",
   "tests": [
     {
```
ErikSchierboom: We've recently added the `"expected"` property to each test entry. See https://github.com/exercism/research_experiment_1/blob/master/exercises/ruby-1-a/.meta/config.json#L9 Could you add the `"expected"` value to all entries? FYI: this is how expected `string` values should be encoded: https://github.com/exercism/research_experiment_1/blob/master/exercises/ruby-1-a/.meta/config.json#L27

exercises/ruby-1-b/ruby_1_b_example.rb
```diff
@@ -1,12 +1,95 @@
...
-def word_with_most_vowels(sentence)
-  words = sentence.split(" ")
-  word_vowels = words.map do |word|
-    [
-      word,
-      word.downcase.chars.select{|l|VOWELS.include?(l)}.count
-    ]
-  end
-  word_vowels.sort_by {|w,c| -c}.first.first
+class Library
```
ErikSchierboom: Nice example. Could you try to create three different examples that all use different approaches? Similar to what `ruby-1-a` did? https://github.com/exercism/research_experiment_1/tree/master/exercises/ruby-1-a

exercises/ruby-1-b/README.md
```diff
@@ -4,17 +4,27 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
 
-Your task is to determine which word in a sentence has the most vowels (a, e, i, o, u). A sentence is defined as a [Ruby string][docs-string] consisting of one or more words. A word consists of one or more upper- or lowercase letters. Words are separated from each other by a single space character.
+You are given a collection of library books, and your job is to import them into
+the library's catalog system. The books are in the constant `Library::BOOKS`, and
+each book has the following columns:
+title, ISBN, author, and publishing year.
 
-Your job is to define a function named `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+You need to write the logic for finding a book, adding new copies to the library's
+inventory, lending a book, and returning a book
```
ErikSchierboom:
```suggestion
inventory, lending a book, and returning a book.
```

exercises/ruby-1-b/README.md
```diff
@@ -4,17 +4,27 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
-For example:
-
-```ruby
-word_with_most_vowels("Hi there aunt jeanie")
-// Returns: "jeanie"
+A book should be findable with it's ISBN
+```
+book = Library.new.find("9780191604928")
+```
+And it should have getters for the attribute values provided in `Library::BOOKS`
```
ErikSchierboom:
```suggestion
And it should have getters for the book's attributes:
```

exercises/ruby-1-b/README.md
```diff
@@ -4,17 +4,27 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
 ```
+book.author
+=> "Mark Twain"
 
-[docs-string]: https://ruby-doc.org/core-2.4.1/String.html
+book.title
+=> "The Adventures of Tom Sawyer"
```
ErikSchierboom: This is slightly more consistent with the [ruby-1-a README](https://github.com/exercism/research_experiment_1/tree/master/exercises/ruby-1-a):

```suggestion
// Returns: "The Adventures of Tom Sawyer"
```

exercises/ruby-1-b/README.md
```diff
@@ -4,17 +4,27 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
-word_with_most_vowels("Hi there aunt jeanie")
-// Returns: "jeanie"
+A book should be findable with it's ISBN
+```
+book = Library.new.find("9780191604928")
+```
+And it should have getters for the attribute values provided in `Library::BOOKS`
 ```
+book.author
+=> "Mark Twain"
```
ErikSchierboom: This is slightly more consistent with the [ruby-1-a README](https://github.com/exercism/research_experiment_1/tree/master/exercises/ruby-1-a):

```suggestion
// Returns: "Mark Twain"
```

exercises/ruby-1-b/README.md
```diff
@@ -4,17 +4,27 @@ This is Part 1 of our research into how people write Ruby differently. This exer
...
-Your job is to define a function named `word_with_most_vowels`, which takes a sentence, and returns the word with the most vowels.
+You need to write the logic for finding a book, adding new copies to the library's
+inventory, lending a book, and returning a book
 
-For example:
-
-```ruby
-word_with_most_vowels("Hi there aunt jeanie")
-// Returns: "jeanie"
+A book should be findable with it's ISBN
```
ErikSchierboom:
```suggestion
A book should be findable with it's ISBN:
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,19 @@
...
+    Programming Ruby: The Pragmatic Programmers' Guide;9780974514055;David Thomas;2004
+    Pride and Prejudice by Jane Austen;9781986431484;Jane Austen;2018
+    To Kill a Mockingbird;9780446310789;Harper Lee;1988
+    Cosmicomics;9780330319089;Italo Calvino;1969
+    The Lord of the Rings;9780544003415;J. R. R. Tolkien;2012
+    Lord of the Flies;9780140283334;William Golding;1999
+    1984: A Novel;9780451524935;George Orwell;2009
+  DATA
 
+  def find
```
ErikSchierboom: Aren't there more methods the student has to implement? Like `borrow`, `return` and such. If so, maybe also add them?

exercises/ruby-1-b/ruby_1_b_example.rb
ErikSchierboom: I'm wondering if this example is perhaps already took bit for being solved in 10-15 minutes? There is already lots going on. We can perhaps strip things out a bit and move additional requirements to the `ruby-2-b` part?

exercises/ruby-1-b/.meta/config.json
goalaleo: I added the key for most of the tests, but I wasn't sure how I should represent return values which are objects. Should I maybe change the methods to return simpler data types like a boolean? e.g. there's no real reason why `borrow` couldn't return `true` instead of a book object

exercises/ruby-1-b/ruby_1_b_example.rb
goalaleo: I'll create 2 more examples once I'm sure which stuff should be moved to the second part 👍

goalaleo: Thanks for the review! I agree that there is a bit too much stuff in this part. What do you think about this division:

Part 1:
- find
- add_copies
- copies (new method, returns number of copies)

Part 2:
- borrow
- return
- borrowed
- (perhaps something else)

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,35 @@
...
+    The Lord of the Rings;9780544003415;J. R. R. Tolkien;2012
+    Lord of the Flies;9780140283334;William Golding;1999
+    1984: A Novel;9780451524935;George Orwell;2009
+  DATA
+
+  def find(isbn)
+    # start by implementing this Library#find method
+  end
+
+  def add_copies(isbn, count: count)
```
iHiD:
```suggestion
  def add_copies(isbn, count:)
```

exercises/ruby-1-b/README.md
```diff
@@ -4,17 +4,27 @@ This is Part 1 of our research into how people write Ruby differently. This exer
 
 ## Instructions
```
iHiD: This README needs to be a full spec of what needs to happen. People cannot see the tests when solving this exercise. So things like `.published` and the other methods with their meanings and things like `ArgumentError` all need explicitly specifying in here.

exercises/ruby-1-b/ruby_1_b_test.rb
```diff
@@ -1,34 +1,95 @@
...
+  def test_book_has_published
+    assert_equal "1988", @library.find("9780446310789").published
+  end
+
+  def test_add_copies_with_invalid_isbn_raises_argument_error
+    assert_raises(ArgumentError) { @library.add_copies("123456789", count: 10) }
+  end
+
+  def test_add_copies_with_valid_isbn_returns_the_added_count
+    assert_equal 10, @library.add_copies("9780544003415", count: 10)
```
iHiD:
```suggestion
    isbn = "9780544003415"
    @library.add_copies(isbn, count: 10)
    assert_equal 10, @library.find(isbn).num_copies
```

exercises/ruby-1-b/ruby_1_b_test.rb
```diff
@@ -1,34 +1,95 @@
...
+  end
+
+  def test_borrowing_when_there_are_no_copies_returns_nil
+    assert_nil @library.borrow("9780544003415")
+  end
+
+  def test_borrowing_a_book_after_copies_have_been_added_returns_the_book
```
iHiD: It's not clear to me what the `borrow` and `return` methods should actually do.

I'd also suggest we don't have a method called return, as this is going to clash when used internally in the class.

kytrinyx:
My biggest concern with the exercise as it is currently implemented, is that the tests are making assumptions about underlying types in the library / catalog, which constrains how people could approach the problem.

I’m also unsure of the different types of implementations someone **could** make that would showcase the difference in their background. For example, how could someone solve this in a more functional manner? Or how might someone’s background change which data structures they choose to use?

In order to ensure that we are not constraining solutions I think there needs to be a clear separation between manipulating the library or catalog (importing books, borrowing/returning books), and reporting on the current state of the library.

I think we should discuss the shape of Part 1 and Part 2 for this exercise and agree on what the exercise should consist of before continuing down the path of getting the test suite right.

What do you all think?

goalaleo:
Ok, I'll try to rearticulate the feedback to make sure I understood it correctly:

The main issues are that:
- the tests are too constraining, and don't facilitate different approaches
- it's not clear what the methods should do, and we should check side effects of method calls instead of method calls

What if I change the exercise like this:

Remove tests that expect a book object with attributes getters, and move those methods to the library. So expect `Library#author` method, which is called with an isbn like 
```ruby
library.author("9780718198916")
=> "Plato"
```
And same for other book attr_readers.

Don't enforce return values for methods `borrow` and `return` (name to be changed). Instead check that calling them has a side effect of changing some state, which can be checked with methods like
```
library.borrowed
=> 3
library.num_copies
=> 10
```

iHiD:
@goalaleo Thanks for being so responsive to the feedback :)

> the tests are too constraining, and don't facilitate different approaches

Yeah - this is the one I'm most generally worried about with the exercise. I'm struggling to think about how this could be approached in a wide variety of different ways. This is true for lots of exercises tbh. But I might just be not considering the different ways that people think (which is what the experiment is trying to work out in a way! :))

> it's not clear what the methods should do, and we should check side effects of method calls instead of return values

Correct. I think shaping the public API of library to imagine you're someone ringing up asking for some information, or borrowing/returning a book etc, and then thinking about how the library would respond to them.

---

So here's a version that I think, while its still going to have a generally similar implementation will probably test someone's focus on OOP vs primitive obsession to the extent that it generates some interesting results.

```ruby
class Library

  # Title;ISBN;Author;Published
  CATALOG_DATA = <<~DATA
    The Adventures of Tom Sawyer;9780191604928;Mark Twain;2007
    Republic;9780718198916;Plato;2012
    Programming Ruby: The Pragmatic Programmers' Guide;9780974514055;David Thomas;2004
    Pride and Prejudice by Jane Austen;9781986431484;Jane Austen;2018
    To Kill a Mockingbird;9780446310789;Harper Lee;1988
    Cosmicomics;9780330319089;Italo Calvino;1969
    The Lord of the Rings;9780544003415;J. R. R. Tolkien;2012
    Lord of the Flies;9780140283334;William Golding;1999
    1984: A Novel;9780451524935;George Orwell;2009
  DATA

  def initialize
    @catalog = CATALOG_DATA.each_line(chomp: true).with_object({}) do |line, obj|
      obj[isbn] = Book.new(*l.split(';'))
    end
  end

  def add_stock!(isbn, count)
    find_book(isbn).add_stock!(count)
  end

  def borrow!(isbn)
    find_book(isbn).borrow!
  end

  def put_back!(isbn)
    find_book(isbn).put_back!
  end

  def book_info(isbn)
    book = find_book(isbn)
    {
      title: book.title,
      author: book.author,
      publication_year: book.publication_year,
      num_in_stock: book.num_in_stock
    }
  end

  def book_in_stock?(isbn)
    find_book(isbn).in_stock?
  end

  def books_with_stock
    catalog.select(&:in_stock?).map(&:isbn)
  end

  def total_books_in_stock
    catalog.sum(&:num_in_stock)
  end

  private
  attr_reader :catalog

  def find_book
    book = catalog[isbn]
    raise ArgumentError.new("No book with isbn: #{isbn}") unless book

    book
  end

  Book = Struct.new(:title, :isbn, :author, :publication_year) do
    def initialize(*args)
      super

      @num_owned = 0
      @num_loaned = 0
    end

    def add_stock!(count)
      self.num_owned += count
    end

    def loan!
      self.num_loaned += 1
    end

    def put_back!
      self.num_loaned -= 1
    end

    def num_in_stock
      (num_owned - num_loaned)
    end

    def in_stock?
      num_in_stock > 0
    end

    private
    attr_accessor :num_owned, :num_loaned
  end

  private_constant :Book
end
```

For each public method of Library we'd have a test with a test about how that affects the library, or it's value in the specific cases where we're asking the library for information. These tests would build on each other, so many look like:
```ruby
def test_add_stock_changes_count_correctly
  library = Library.new
  library.add_stock!('9781986431484', 4)
  assert_equal 4, library.num_in_stock('9781986431484')
end

def test_borrow_changes_count_correctly
  library = Library.new
  library.add_stock!('9781986431484', 4)
  library.borrow('9781986431484')
  assert_equal 3, library.num_in_stock('9781986431484')
end

def test_puts_back_changes_count_correctly
  library = Library.new
  library.add_stock!('9781986431484', 4)
  library.borrow('9781986431484')
  library.put_back('9781986431484')
  assert_equal 4, library.num_in_stock('9781986431484')
end
```

The extension for 2-b could then be that we want to track **who** borrows each book too, which I think would add a nice intersection of part-1 code while also adding interesting new functionality. So I think this would bode well for that.

I've just written all that in GitHub's editor, so it's probably got typos etc, but maybe that general vibe would move us forward?

iHiD:
Update: I've just spent some time with Katrina discussing this a little more. I've updated my example above with an amended version. The changes are based on us just thinking about how to force people down the fewest enforced routes and give people the most freedom. One other thing that came out of this is that we'd rather **not** test for invalid ISBNs as this enforces patterns even more (for example my `find_book` method is a direct result of that test, and I think it's the route most people would choose to take.

Here to answer questions or discuss this more! :)

exercises/ruby-1-b/ruby_1_b_test.rb
kotp: Name options:

- `check_out`
- `check_in`

I went through the English thesaurus I had available, and remit, lend, borrow, all had their charms, but quirks.  I think the above is probably the clearest names, avoiding the `return` that going to be problematic in Ruby.

goalaleo:
Thanks for the feedback @iHiD :) ! I'll update the exercise based on your suggestions in the evening.

Just to be sure that I understood your comment on invalid ISBNs:
So should I remove the test that asserts that an ArgumentError is raised? And your implementation of the method is an example of what happens if we do assert it, and in your opinion we shouldn't enforce students to validate the input?

iHiD:
> Thanks for the feedback @iHiD :) ! I'll update the exercise based on your suggestions in the evening.

Great. Thanks!

>So should I remove the test that asserts that an ArgumentError is raised? And your implementation of the method is an example of what happens if we do assert it, and in your opinion we shouldn't enforce validating the input?

Yes. Correct. My implementation still had it but after discussing with Katrina we realised that it pushes someone in quite a firm direction. For example, it would be more interesting to see if people:
- Don't handle the case
- Use nils
- Use exceptions

And whether that is indicative of their background :)

goalaleo:
@iHiD I updated the exercise and tests - let me know what you think! I'll update the config and README once the example has been approved

exercises/ruby-1-b/ruby_1_b_test.rb
```diff
@@ -1,34 +1,78 @@
...
+    @library.add_stock!("9780330319089", 1)
+    @library.borrow!("9780330319089")
+    @library.put_back!("9780330319089")
+    assert_equal 1, @library.lookup_stock("9780330319089")
+  end
+
+  def test_books_in_stock
+    isbns = ["9780544003415", "9780140283334", "9780451524935"]
+    isbns.each { |isbn| @library.add_stock!(isbn, 1) }
+    assert_includes @library.books_in_stock, "9780544003415"
```
iHiD: This could probably be `assert_equal ibsns.sort, @library.books_in_stock.sort` and then delete the next few lines.

goalaleo: Good idea 👍 I'll update it

goalaleo:
As the example has received the green light, I'll update the config and README next. I'd be happy to also receive suggestions on different approaches for the other examples.

ErikSchierboom:
Good to hear @goalaleo! We're getting there!

goalaleo:
@iHiD do you think that we could leverage other ruby contributors/maintainers for more organic alternative approaches? I'm a bit concerned that they'd look somehow "artificially" different if I try to write a solution that should explicitly differ from the first example.

kotp:
> @iHiD do you think that we could leverage other ruby contributors/maintainers for more organic alternative approaches? I'm a bit concerned that they'd look somehow "artificially" different if I try to write a solution that should explicitly differ from the first example.

Please do ask on the Slack [Ruby channel](https://exercism-team.slack.com/archives/CARVB5V0R), @goalaleo as that is where a few mentors will see it.  I believe the maintainers will see the request here, but we can also mention it on Slack [Ruby Maintainers](https://exercism-team.slack.com/archives/CC39FFMU0) channel to ping anyone that may not have seen it from the maintainers pool.

iHiD:
The different approaches are just to prove that the problem can be solved in different interesting ways, and that therefore it's valid. If you can find those three ways, then that's all that matters for this part.

goalaleo:
@ErikSchierboom the README and config.json are now up to date. Let me know if you have any more feedback. This is only missing the 2 other example approaches

P.S. What's the policy for @mentioning people - should I do it?

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,7 +1,7 @@
 class Library
 
   # Title;ISBN;Author;Published
-  BOOKS = <<~DATA
+  CATALOG_DATA = <<~DATA
```
kotp: Is `_DATA` redundant here?  Any catalog I have ever seen contains data, so that could be assumed.

kotp:
> @ErikSchierboom the README and config.json are now up to date. Let me know if you have any more feedback. This is only missing the 2 other example approaches
> 
> P.S. What's the policy for @mentioning people - should I do it?

Yes, you should.

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+  DATA
+
+  def lookup_title(isbn)
+  end
+
+  def lookup_author(isbn)
+  end
+
+  def lookup_publication_year(isbn)
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#lookup_publication_year method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+
+  def add_stock!(isbn, count)
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#add_stock! method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
 
+  def lookup_stock(isbn)
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#lookup_stock method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+
+  def borrow!(isbn)
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#borrow! method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+
+  def put_back!(isbn)
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#put_back! method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+
+  def book_in_stock?(isbn)
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#book_in_stock? method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+
+  def books_in_stock
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#books_in_stock method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -1,4 +1,45 @@
...
+
+  def total_books_in_stock
+  end
```
ErikSchierboom:
```suggestion
    raise NotImplementedError, 'Please implement the Library#total_books_in_stock method'
  end
```

exercises/ruby-1-b/ruby_1_b.rb
iHiD: I think that `Catalog` implies that it's an actual catalog, but it's not - it's a CSV. And because I'd rather not encourage the use of the CSV library, data feels like a more agnostic choice. I'm happy for an alternative word, or even for this just to be called `DATA`, but I think we shouldn't imply that it's a rich-object.

ErikSchierboom:
Thanks @goalaleo!

exercises/ruby-1-b/ruby_1_b.rb
```diff
@@ -43,5 +43,6 @@ def books_in_stock
   end
 
   def total_books_in_stock
+    raise NotImplementedError, 'Please implement the Library#total_books_in_stock method'
```
kotp: A little bit late for this... but...

```ruby
raise NotImplementedError, "Please implement the #{self.class}##{__method__} method."
```

goalaleo:
Thank you everyone for support!

PR #59, opened 2020-01-29
[Ruby] Add expected value and update cmd

ErikSchierboom:


exercises/ruby-1-a/ruby_1_a_test.rb
```diff
@@ -23,8 +23,8 @@ def test_two_items
 
   def test_list_items
     cart = ShoppingCart.new
-    cart.add(:STARIC)
     cart.add(:STAPOT)
```
ErikSchierboom: I've changed the order as otherwise the next test (alphabetic listing is returned) does not test any new behavior.

PR #66, opened 2020-01-30
ruby: Ensure instructions are complete for ruby-1-a

kytrinyx:
I played with the Ruby 1 a exercise, and I really like it. It seems like it has a really good scope and level of complexity.

There were a few things that were unclear as I was working through it:

- the SKU should be a Symbol rather than a String. This was not
specified, and the example showed a constant, which doesn't tell the
research subject what the type is.
- the price list shows a combination of integers and floats. The
instructions state that the total amount should have 2 decimal places,
but the example shows a return value of an integer with no decimal
places.
- the example shows the item names in lowercase, whereas the actual
table shows the names capitalized.
- the README doesn't specify that the `item_list` requires the names
to be unique.

I also updated the tests to assert against floats consistently, even
though Ruby considers 1 == 1.0.

/cc @pvcarrera as original author ✨

### DIFF
--- a/exercises/ruby-1-a/README.md
+++ b/exercises/ruby-1-a/README.md
@@ -10,28 +10,29 @@ Items are added to the basket using stock-keeping unit (SKU) codes. The items in
 
 | Name      | Price | SKU    |
 |-----------|-------|--------|
-| Potatoes  | 10    | STAPOT |
-| Rice      | 30    | STARIC |
+| Potatoes  | 10.00 | STAPOT |
+| Rice      | 30.00 | STARIC |
 | Coffee    | 14.99 | STACOF |
-| Newspaper | 2.99  | MEDNEW |
+| Newspaper |  2.99 | MEDNEW |
+
 
 You need to define a `ShoppingCart` class, with three public methods:
-- `add`: Takes a SKU and adds it to the cart. The SKU is comprised of the first three upcased chars of the product type + the first three upcased chars of the product name.
+- `add`: Takes a SKU and adds it to the cart. The SKU is a `Symbol` comprised of the first three uppercase letters of the product type + the first three uppercase letters of the product name.
 - `total_amount`: The value of the items in the cart as a number with two decimal places (e.g. 3.99).
-- `items_list`: A list of the names of the items in the cart, ordered alphabetically, separated by a comma and a space.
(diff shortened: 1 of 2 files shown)


pvcarrera:
Awesome! Thank you for all the help with the exercise, It looks much better than when I started 😄 
This weekend I'm planing on working on the second half which is going to include discounts and maybe dealing with float numbers or currencies.

exercises/ruby-1-a/README.md
```diff
@@ -13,25 +13,26 @@ Items are added to the basket using stock-keeping unit (SKU) codes. The items in
 | Potatoes  | 10    | STAPOT |
 | Rice      | 30    | STARIC |
 | Coffee    | 14.99 | STACOF |
-| Newspaper | 2.99  | MEDNEW |
+| Newspaper |  2.99 | MEDNEW |
```
iHiD: Is this extra whitespace intentional? I think it will be stripped in the UI anyway (as double spaces get rendered as one in HTML)

exercises/ruby-1-a/README.md
```diff
@@ -13,25 +13,26 @@ Items are added to the basket using stock-keeping unit (SKU) codes. The items in
 | Potatoes  | 10    | STAPOT |
```
ErikSchierboom: @kytrinyx As we're now using numbers with two decimals everywhere else, should we also be using decimals for the whole numbers too?

```suggestion
| Potatoes  | 10.00    | STAPOT |
```

exercises/ruby-1-a/README.md
kytrinyx: I did it to line up the decimal point. I don't feel strongly. I originally added decimals to the integers as well, but backed it out after a discussion (can't remember why now...)

kytrinyx: Oh, and Erik changed it back. I do like it better that way :)

exercises/ruby-1-a/ruby_1_a_test.rb
```diff
@@ -5,20 +5,20 @@
 
 class ShoppintCartTest < Minitest::Test
```
iHiD:
```suggestion
class ShoppingCartTest < Minitest::Test
```

kotp:
Thank you @kytrinyx and @iHiD .

PR #74, opened 2020-01-30
go: Tweak go-1-a

kytrinyx:
This updates the README to say 'function name' rather than 'signature'
when describing just the name of Description().

It also tweaks the stub file in two ways:

1. it edits the function comment to be a bit more go-ish (sorry), and
2. it renames the function parameter to `number` rather than `amount`,
since this is a countable, not a measurable.

Lastly, I updated the output from the test failures to make it a bit
more obvious what the 'got' vs the 'want' is (got and want are the
canonical terms used by the Go team, so I went with that).

    --- FAIL: TestTwelve (0.00s)
        go_1_a_test.go:40: Describe(12)
               got: "12"
              want: "1 dozen or 12".

### DIFF
--- a/exercises/go-1-a/go_1_a_test.go
+++ b/exercises/go-1-a/go_1_a_test.go
@@ -37,6 +37,6 @@ func TestSixtyFive(t *testing.T) {
 func AssertDescribe(t *testing.T, input int, expected string) {
 	actual := Describe(input)
 	if actual != expected {
-		t.Errorf("Describe(%d) = %q, expected %q.", input, actual, expected)
+		t.Errorf("Describe(%d)\n   got: %q\n  want: %q.", input, actual, expected)
 	}
 }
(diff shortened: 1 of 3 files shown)


exercises/go-1-a/go_1_a_test.go
```diff
@@ -37,6 +37,6 @@ func TestSixtyFive(t *testing.T) {
 func AssertDescribe(t *testing.T, input int, expected string) {
 	actual := Describe(input)
 	if actual != expected {
-		t.Errorf("Describe(%d) = %q, expected %q.", input, actual, expected)
+		t.Errorf("Describe(%d)\n   got: %q\n  want: %q.", input, actual, expected)
```
ErikSchierboom: This is nice! We should probably also do this for the other Go test files at some point.

ErikSchierboom: I can handle that.

PR #88, opened 2020-01-31
[Ruby] 1 A: Replaces constant with symbol for SKU

kotp:
Ref: #66


kotp:
Looks like this is being addressed... apologize for the noise.

PR #99, opened 2020-02-06
[Ruby] Extension for the shopping cart exercise

pvcarrera:
I've been doing some work on the extension for the shopping cart exercise and I'd like to get some feedback before continuing. 

As you can see I added the `scan_coupon` method and two types of discounts `half price` and `fix`. What do you think about the scope and level of complexity?  As you can see, I already have three more tests in mind that will increase the scope and complexity but I'm not sure where we want to stop.

Any other suggestions will be more than welcome, thanks!

### DIFF
--- a/exercises/ruby-2-a/.meta/config.json
+++ b/exercises/ruby-2-a/.meta/config.json
@@ -1,0 +1,78 @@
+{
+  "solution_files": ["ruby_1_a.rb"],
+  "test_file": "ruby_1_a_test.rb",
+  "tests": [
+    {
+      "test": "test_initial_total_amount",
+      "name": "Initial total amount",
+      "cmd": "ShoppingCart.new.total_amount",
+      "expected": "0"
+    },
+    {
+      "test": "test_one_item",
+      "name": "One item",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STAPOT)\ncard.total_amount",
+      "expected": "10"
+    },
+    {
+      "test": "test_two_items",
+      "name": "Two items",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STAPOT)\ncart.add(:STARIC)\ncart.total_amount",
+      "expected": "40"
+    },
+    {
+      "test": "test_list_items",
+      "name": "List items",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STAPOT)\ncart.add(:STARIC)\ncart.items_list",
+      "expected": "\"Potatoes, Rice\""
+    },
+    {
+      "test": "test_items_ordered_alphabetically",
+      "name": "List items alphabetically",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STARIC)\ncart.add(:STAPOT)\ncart.items_list",
+      "expected": "\"Potatoes, Rice\""
+    },
+    {
+      "test": "test_duplicate_items_sums_price",
+      "name": "Total amount works with duplicates",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STARIC)\ncart.add(:STACOF)\ncart.add(:STACOF)\ncart.total_amount",
+      "expected": "59.98"
+    },
+    {
+      "test": "test_duplicate_items_uniqs_list",
+      "name": "Items list does not contain duplicates",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STARIC)\ncart.add(:MEDNEW)\ncart.add(:MEDNEW)\ncart.items_list",
+      "expected": "\"Newspaper, Rice\""
+    },
+    {
+      "test": "test_fix_price_coupon",
+      "name": "Apply fix price coupon, amount = 10",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STAPOT)\ncart.add(:MEDNEW)\ncart.add(:MEDNEW)\ncart.scan_coupon(:FIX_TEN)\ncart.total_amount",
+      "expected": "5.98"
+    },
+    {
+      "test": "test_items_list_does_not_include_discounts",
+      "name": "Discounts are not included on the items list",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:MEDNEW)\ncart.scan_coupon(:FIX_TEN)\ncart.items_list",
+      "expected": "\"Newspaper\""
+    },
+    {
+      "test": "test_product_discount_coupon",
+      "name": "Apply product discount coupon: Potatoes are half price",
+      "cmd": "cart = ShoppingCart.new\ncart.add(:STAPOT)\ncart.add(:STAPOT)\ncart.add(:MEDNEW)\ncart.scan_coupon(:HALF_STAPOT)\ncart.total_amount",
+      "expected": "12.99"
+    },
+    {
+      "test": "test_coupon_does_not_apply",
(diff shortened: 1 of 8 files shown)


iHiD:
I actually think what you have there now is great. I wouldn't add the extra tests. It will mean fiddling with existing code enough that it will give us the sort of data I think we want. So I'd just tidy up what you've done here.

exercises/ruby-2-a/ruby_2_a_test.rb
```diff
@@ -1,9 +1,120 @@
...
+  def test_minimum_purchase_amount
+    skip
+  end
 
-# Common test data version: 1.2.0 4fc1acb
-class StubbedTest < Minitest::Test
-  def test_passes
-    assert true
+  def test_buy_two_get_one_free_coupon
+    skip
```
kotp: I am assuming this is till a work in progress?  Otherwise remove the skip tests that are similar to this one.

ErikSchierboom: Yes. As @iHiD said: "I wouldn't add the extra tests." So I think these can be skipped.

kotp: Right, but should they be skipped and here, or gone and therefore no need to skip?

ErikSchierboom: Oh that was a bad choice of words! I meant: removed :)

pvcarrera:
Ok I cleaned up the tests and the example, added the README and the `.meta`. I'd say now it's ready to go through the normal code review process 😄

exercises/ruby-2-a/README.md
```diff
@@ -1 +1,31 @@
...
+## Instructions
+
+Your task is to extend the existing shopping cart with a new functionality. Once the extension is added the shopping cart should be able to read and apply discount coupons to the current items. The supported coupons are:
+
+| Type                 | SKU         | Rule                                             |
+| -------------------- | ----------- | ------------------------------------------------ |
+| Fixed price discount | FIX_TEN     | Discounts 10.00 off the total amount of the cart |
+| Product discount     | HALF_STAPOT | Reduces all STAPOT items to half price           |
+
+You task is to define a new `ShoppingCart#scan_coupon` method which takes a coupon SKU and adds it to the cart. The SKU is a `Symbol` comprised of the first three uppercase letters of the discount type + the SKU of the relevant product. In case the coupon type is fixed price, the second part of the SKU is the discounted amount in uppercase letters. Coupons are unique and can be applied only once.
```
kotp: This is the first use of `SKU` here, should we define that again?  The meaning of SKU, not what it represents as spelled out here, as to what it is comprised of, but what it stands for.

exercises/ruby-2-a/README.md
```diff
@@ -1 +1,31 @@
...
+
+Please bear in mind that the `total_amount` can't be a negative number, the minimum value for the `total_amount` is 0.
```
kotp:
```suggestion
Please bear in mind that the `total_amount` must not be a negative number, the minimum value for the `total_amount` must be 0.
```

exercises/ruby-2-a/README.md
```diff
@@ -1 +1,31 @@
...
+shopping_cart.add(:STAPOT)
+shopping_cart.add(:MEDNEW)
+shopping_cart.add(:MEDNEW)
+shopping_cart.scan_coupon(:FIX_TEN)
+
+shopping_cart.total_amount
+// Returns: 5.98
+```
+
+**Note:** You can use `debug("some test")` to output debugging information for yourself.
```
kotp:
```suggestion
**Note:** You may use `debug("some test")` to output debugging information for yourself.
```

exercises/ruby-2-a/.meta/config.json
```diff
@@ -0,0 +1,78 @@
+{
+  "solution_files": ["ruby_1_a.rb"],
```
ErikSchierboom:
```suggestion
  "solution_files": ["ruby_2_a.rb"],
```

exercises/ruby-2-a/.meta/config.json
```diff
@@ -0,0 +1,78 @@
...
+  "test_file": "ruby_1_a_test.rb",
```
ErikSchierboom:
```suggestion
  "test_file": "ruby_2_a_test.rb",
```

exercises/ruby-2-a/ruby_2_a.rb
```diff
@@ -1 +1,19 @@
-TODO
\ No newline at end of file
+# frozen_string_literal: true
```
ErikSchierboom: The second part of the research experiment has its starting points (of which should be multiple) in a different location: `.meta/start/{1,2,3}/ruby_2_a.rb`.  This means that this file can be removed and that the [example implementations to part 1](https://github.com/exercism/research_experiment_1/tree/master/exercises/ruby-1-a) should be used as  `.meta/start/1/ruby_2_a.rb`, `.meta/start/2/ruby_2_a.rb` and `.meta/start/3/ruby_2_a.rb`. To these files, the `scan_coupon` stub should be added, similar to how it is defined below.

See the [csharp-2-b example](https://github.com/exercism/research_experiment_1/tree/master/exercises/csharp-2-b/.meta/start). 

Let me know if this is not clear enough.

exercises/ruby-2-a/README.md
```diff
@@ -1 +1,31 @@
-TODO
+## Introduction
+
+This is Part 2 of our research into how people write Ruby differently. This exercise should take 10-15 minutes for a programmer who is proficient in Ruby. There is no right or wrong way to approach it - just do what feels the most natural to you. Feel free to use books, Google or Stack Overflow, just like you would if you were programming normally, but please don't actively look for a solution to the problem as this will negatively affect the research.
+
+## Instructions
+
+Your task is to extend the existing shopping cart with a new functionality. Once the extension is added the shopping cart should be able to read and apply discount coupons to the current items. The supported coupons are:
```
ErikSchierboom: I would suggest to also include a description of the existing code here, to have the student get an idea what the current code does. See [this C# example](https://github.com/exercism/research_experiment_1/tree/master/exercises/csharp-2-b#instructions).

exercises/ruby-2-a/README.md
ErikSchierboom: If we apply this suggestion (to me as a non-native speaker it looks the same :)), we should also update the [ruby-1-a readme](https://github.com/exercism/research_experiment_1/blob/master/exercises/ruby-1-a/README.md).

iHiD: I actually think this should be `can`, as we're telling someone that something is possible, rather (can) than something is allowed (may).

exercises/ruby-2-a/ruby_2_a.rb
pvcarrera: For the second part I only have one example. Do we need one example for each start?

ErikSchierboom: No, one example is fine!

pvcarrera:
@ErikSchierboom following your advice, I've added some new commits to improve the README and add the `start` exercises. Let me know what you think.

ErikSchierboom:
@pvcarrera Don't forget to add yourself to the maintainers.json file (see https://github.com/exercism/research_experiment_1/pull/111/files for an example).

PR #104, opened 2020-02-13
[Docs] Add ErikSchierboom as maintainer

ErikSchierboom:


config/maintainers.json
```diff
@@ -1,4 +1,20 @@
...
+    {
+      "github_username": "ErikSchierboom",
+      "name": "Erik Schierboom",
+      "link_text": "My blog",
+      "link_url": "http://www.erikschierboom.com/",
+      "avatar_url": "https://nl.gravatar.com/userimage/27030165/76c834a4087478d5f9ff6ff8c6c1429e.jpg",
+      "bio": "I am a developer with a passion for learning new languages. My favorite languages are F# and Kotlin.",
+      "contributions": [
+        "csharp/test-runner",
+        "csharp/research-exercise",
```
iHiD: Maybe just change to `exercise` as that's the only option here?

ErikSchierboom: My reasoning was that we might want to use the same format later for other repositories. You could then either have logic that would convert an exercise in the research_experiment_1 repo to mean a research exercise, or else we add the prefix. My other thought was that we could separate "csharp/concept-exercise" and "csharp/practice-exercise" later, so this makes it consistent with that.

iHiD: I suspect that we'll move to having contributors listed in the actual exercise config.json files rather than have an uber-file for it. But it's too early to tell.

Let's merge this for now and it's an easy change either way later :+1:

ErikSchierboom: 👍

PR #110, opened 2020-02-14
[python] Part 2 files

yawpitch:
As per discussion with @ErikSchierboom  the README.md and .meta/config.json are not yet done.


exercises/python-2-b/example.py
```diff
@@ -1 +1,77 @@
...
+
+def receive(transmission: str) -> str:
+    """
+    Convert a NATO code word transmission to a message.
+    """
+    return "".join(map(NATO_TO_ALPHANUM.get, transmission.split()))
+
+ALPHANUM = "".join(ALPHANUM_TO_NATO.keys())
+CAESAR = str.maketrans(ALPHANUM, ALPHANUM[-3:] + ALPHANUM[:-3])
+RASEAC = {v: k for k, v in CAESAR.items()}
```
ErikSchierboom: LOL!

exercises/python-2-a/python_2_a_test.py
```diff
@@ -1 +1,96 @@
...
+"""
+import unittest
+import string
+from itertools import chain
+
+from python_2_a import compress, decompress
+
+
+class TestCompress(unittest.TestCase):
+    def assertCanCompressAndDecompress(self, string, expect, slug):
```
ErikSchierboom: This is the only thing I'm not sure of in this PR. It means that all existing tests are now no longer valid. I'm not sure if that is an issue, but I'd like to check with @iHiD.

yawpitch: Well, to be fair, the previous tests were doing the exact same thing, they just had a `_decompress` implementation built into them which was verifying that the student's compression could roundtrip ... now it's verifying the exact same thing, only they've got to provide the working implementation.

ErikSchierboom: Right! Then it's perfectly okay I think. I'll merge this and add the other documents later.

ErikSchierboom:
Super sweet to have Python close to done now!

PR #134, opened 2020-04-25
[Python] Fix Typos

yawpitch:
A couple of tests had transposed function names. Thanks to user Yoni (ynfle) for catching these.


iHiD:
Thanks @yawpitch. So FYI, this auto-deploys when merged (this bit lives in the website not the test runner). However, it won't update for people who have already started the exercises (in the same way people need to click the "update to latest version" for normal tracks - but the button isn't there). I can run a script to sort it, but just for your knowledge of the internals.

PR #135, opened 2020-04-26
[Common Lisp] Add Common Lisp Exercise 1A

TheLostLambda:
Hello! This is a first draft of things and I'm still working on coming up with a second challenge. I'd love to get some feedback on this one though! 

I think there are a ton of different ways to approach the problem and it should be very easy to extend in part 2


verdammelt:
I just solved this relatively quickly and one feedback I have is that it might be better for flow if the tests were ordered as: twin, diff-atk, diff-hp, diff-type. (diff-atk, diff-hp in any order). 

That way type-effectiveness is put off until all the rest of the mechanism of battle is resolved.  Sort of the way I'd do it if I were TDDing it.

TheLostLambda:
That's an excellent point about the test order! I think I'll swap that around the way you've suggested. And since the magic time is 10-15 minutes, would you say it took around that amount of  time to solve?

Any other things that you reckon could be improved?

exercises/common-lisp-1-a/.meta/example.lisp
```diff
@@ -0,0 +1,19 @@
...
+(defmethod tm ((a (eql 'water)) (b (eql 'fire))) 2)
+(defmethod tm (a b) (if (eq a b) 1 0.5))
+
+(defmacro attack (a b)
+  `(let ((type-modifier (tm (pokemon-type ,a) (pokemon-type ,b))))
+     (decf (pokemon-hp ,b) (* (pokemon-atk ,a) type-modifier))
+     (when (< (pokemon-hp ,b) 1) (return (pokemon-name ,a)))))
+
+(defun battle (a b)
+  (loop (attack a b) (attack b a)))
```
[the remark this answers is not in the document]
TheLostLambda: Yup! I've definitely had fun abusing Lisp's macro power :smile:

verdammelt:
I don't remember how long it too me. Probably too long :).

I don't think it would have bee much over 10-15 minutes especially if one didn't go in with a clear (but wrong) design in mind :)

exercises/common-lisp-1-a/.meta/example-1.lisp
```diff
@@ -0,0 +1,19 @@
+(defpackage pokemon-battle
```
ErikSchierboom, line 1: There are currently two different example implementations, that do look quite different, which is great! We strive to have three different example implementations, which can then be used as the starting points for the second part of the exercise. Is it feasible to have a third, different example?

exercises/common-lisp-1-a/README.md
```diff
@@ -0,0 +1,59 @@
...
+As an aspiring [Pokémon][pokemon] trainer, you're looking to get some battle
+experience so you can hone your skills. The only problem is, you don't have any
+Pokémon! Being the resourceful trainer you are, you decide to write a simple
+computer program that will allow you to simulate Pokémon battles so you can be
+prepared for the real deal.
+
+### Making Pokémon
+
+In this simplified simulator, every Pokémon has a name, type, attack stat, and
+some HP (hit-points). You should define a function `make-pokemon` for
```
ErikSchierboom, line 12: Would it be possible to define a stub function for `make-pokemon`? This would help the student get started and is also how the other tracks define their stubs.

exercises/common-lisp-1-a/README.md
```diff
@@ -0,0 +1,59 @@
+## Pokémon Battle
```
ErikSchierboom, line 1: To conform to how other tracks structure their readme:

```suggestion
## Instructions
```

exercises/common-lisp-1-a/.meta/example-1.lisp
TheLostLambda: Certainly! I can work on one of those or try to source some ideas from others!

exercises/common-lisp-1-a/README.md
TheLostLambda: We could define a stub for it if that's what's best. The reason I avoided it is that there are technically solutions that won't need to explicitly define that function. Using something like `defstruct` will automatically define that function and some others.

I left it out thinking it would encourage some more diverse solutions, but I can add it if you would like!

exercises/common-lisp-1-a/README.md
TheLostLambda: Done!

TheLostLambda:
@ErikSchierboom I'm assuming that the `tests` field `.meta/config.json` is something required by the site right now? I think the Common Lisp test runner will already dump this info for failed tests (it shows the name, code, expected value, and actual value for a failed test), but I assume that's just not piped up yet?

Regardless, I'm working on writing the `.meta/config.json` for this exercise, I was just curious :)

exercises/common-lisp-1-a/.meta/example-1.lisp
ErikSchierboom: Excellent!

exercises/common-lisp-1-a/README.md
ErikSchierboom: Hmmm, that is a good argument. Let me think on this!

ErikSchierboom:
> I'm assuming that the tests field .meta/config.json is something required by the site right now? I think the Common Lisp test runner will already dump this info for failed tests (it shows the name, code, expected value, and actual value for a failed test), but I assume that's just not piped up yet?

Yes, that is correct. The `.meta/config.json` file is basically adds metadata to the output of the test runner.

ErikSchierboom:
> Additionally, what does the "test" field in .meta/config.json correspond to? It is the "name" in results.json? If so, our runner shows results for every sub-test individually, so there are duplicate names.

Eh yes, test names must be unique. Can that be done?

TheLostLambda:
Yep! I don't think I need to change the test runner either. I just need to simplify the test file itself!

TheLostLambda:
I've pushed a completed `.meta/config.json` with the tweaked test file. The `"name"` of a test in the `results.json` now matches the `"test"` field of `.meta/config.json`!

iHiD:
@TheLostLambda Thanks :) We can align the namings when we do this for v3 :)

TheLostLambda:
> We can align the namings when we do this for v3 :)

@iHiD Will we be using these type of `.meta/config.json` files in v3? The CL test runner will already return all of this information for failing tests, so these files seem a bit redundant and quite error prone.

I suppose not all languages' test runners will be able to output the test code being run, but can that be an option for tracks with runners that can (as it appears in the `"message"` field)?

iHiD:
> @iHiD Will we be using these type of .meta/config.json files in v3? The CL test runner will already return all of this information for failing tests, so these files seem a bit redundant and quite error prone.

Yes. Literally everyone has said this, then changed their mind when they've seen them live in the UI :)

> I suppose not all languages' test runners will be able to output the test code being run, but can that be an option for tracks with runners that can (as it appears in the "message" field)?

Yes. The key is that the test runner doesn't return the info. It doesn't return the test code that's been run and it doesn't return it in a way that doesn't expose the testing language. That's the key. Take for example:
```
  (is (equal "A" (battle (make-test-pokemon "A" 'fire)
                         (make-test-pokemon "B" 'grass))))
```

As a student, I don't want to be exposed to this assertion framework. I want to just see  ... well actually I literally can't read that so I'm not sure what the code being run is and what's been asserted there. I guess the output is "A"? (This has turned into a great example :)) What i want to see as a student is the code that will get run, and the output I'm expected to output, and what actually happened. 

And this is the best way we've found so far to do it.

I talked a little abut this here and we discussed how we might autogenerate them too: https://youtu.be/7im5LPuDVd0?t=310

exercises/common-lisp-1-a/README.md
ErikSchierboom: @iHiD Any thoughts on this?

TheLostLambda:
Hmm. Okay, I suppose that puts me in the "I hate it, but it's what works" camp... Does this then mean vastly simplifying the test runner? What happens to the message field? Is there a PR somewhere for discussing this?

I think it might be worth relying only on the `results.json` of the test runner for the website, but allow the test-runner itself to access a `config.json` if that's how a particular track decides to approach the issue.

I'll save my further pondering / grovelling for a discussion PR if we make one :smile:

ErikSchierboom:
> Does this then mean vastly simplifying the test runner? What happens to the message field? Is there a PR somewhere for discussing this?

I don't think it does. The `message` field _is_ still being displayed if it is present.

iHiD:
Think of the `results.json` as the results, and the `config.json` as the information that explains how the results have been generated.

I think the config.jsons could be pretty readily automated for most languages though.

ErikSchierboom:
While testing the test runner, I found that the `exercises/common-lisp-1-a/pokemon-battle-test.lisp` and `exercises/common-lisp-1-a/pokemon-battle.lisp` don't conform to the slug name, which meant that calling the test runner with the `common-lisp-1-a` slug did not work. These should be renamed to `exercises/common-lisp-1-a/common-lisp-1-a.lisp` and `exercises/common-lisp-1-a/common-lisp-1-a-test.lisp`.

TheLostLambda:
@ErikSchierboom Excellent catch! I've fixed things now!

exercises/common-lisp-1-a/README.md
iHiD: Good idea, but I think @TheLostLambda made a good argument to leave it as it is. Obviously if this was v3 I'd be hard the other way, but for the research I think allowing multiple possible solutions is the most critical variable.

ErikSchierboom: Okay! Let's approve this then :)

ErikSchierboom:
Thanks for doing this @TheLostLambda!

iHiD:
Thanks @TheLostLambda and reviewers! :D

PR #140, opened 2020-07-15
Julia: Add ex1a tic-tac-toe

SaschaMann:
Before I continue, it would be nice if you could check if this exercise meets the requirements.

As mentioned on Slack, the 2nd part would involve implement an "AI" that selects the next move according to the following rules:

- If the non-AI player can win the next turn, block them.
- If there's no way to block the other player, either because they don't have two of their marks in a row or because they have two winning positions, place the mark "randomly" (the tests won't check if the mark has actually been placed randomly, but instead will check if it returns one of the valid positions).

@ErikSchierboom @iHiD

### DIFF
--- a/exercises/julia-1-a/.meta/example.jl
+++ b/exercises/julia-1-a/.meta/example.jl
@@ -6,8 +6,9 @@ function gamestate(bs)
     nₒ = count(x -> x == "O", b)
     nₓ = count(x -> x == "X", b)
     # Not taking turns in order 
-    nₒ > nₓ && throw(DomainError(b, "Invalid board: did not take turns in order"))
-    abs(nₒ - nₓ) > 1 && throw(DomainError(b, "Invalid board: did not take turns in order"))
+    if nₒ > nₓ || abs(nₒ - nₓ) > 1
+        throw(DomainError(b, "Invalid board: did not take turns in order"))
+    end
 
     wins = 0
 
--- a/exercises/julia-1-a/.meta/example.jl
+++ b/exercises/julia-1-a/.meta/example.jl
@@ -1,0 +1,36 @@
+function gamestate(bs)
+    # Parse to Matrix
+    b = hcat(split.(split(bs, "\n"), "")...)
+
+    # Invalid boards
+    nₒ = count(x -> x == "O", b)
+    nₓ = count(x -> x == "X", b)
+    # Not taking turns in order 
+    nₒ > nₓ && throw(DomainError(b, "Invalid board: did not take turns in order"))
+    abs(nₒ - nₓ) > 1 && throw(DomainError(b, "Invalid board: did not take turns in order"))
+
+    wins = 0
+
+    for i in 1:3
+        # Row victory
+        wins += b[i, 1] != " " && (b[i, 1] == b[i, 2] == b[i, 3]) ? 1 : 0
+        
+        # Col victory
+        wins += b[1, i] != " " && (b[1, i] == b[2, i] == b[3, i]) ? 1 : 0
+    end
+
+    # Diagonal victories
+    wins += b[1, 1] != " " && (b[1, 1] == b[2, 2] == b[3, 3]) ? 1 : 0
+    wins += b[1, 3] != " " && (b[1, 3] == b[2, 2] == b[3, 1]) ? 1 : 0
+
+    if wins == 1
+        return :win
+    elseif wins > 1
+        throw(DomainError(b, "Invalid board: more than one win"))
+    end
+
+    # Draw
+    " " ∉ b && return :draw
+
+    :ongoing
+end
--- a/exercises/julia-1-a/README.md
+++ b/exercises/julia-1-a/README.md
@@ -1,0 +1,98 @@
+## Instructions
+
+Implement your own [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) game.
+(_You may also know the game as "noughts and crosses" or "Xs and Os"._)
+
+The games is played on a 3x3 grid.
+Players take turns to place `X`s and `O`s on the grid.
+The game ends when one player has won by placing three of marks in a row, column, or along a diagonal of the grid, or when the entire grid is filled up.
+
+Without loss of generality we will assume that `X` starts.
+
+At first, you will create create a function `gamestate` that takes the game board represented a string as its sole argument and returns a Symbol describing the state of the game.
+The board will be given as a 3-line string containing `X`s, `O`s and ` `s:
+
+```julia
+board = "X  \n XO\nOX "
+```
+
+represents the following board:
+
+```
+   |   |   
+ X |   |   
+___|___|___
+   |   |   
+   | X | O 
+___|___|___
+   |   |
+ O | X |  
+   |   |
+```
+
+There are 3 potential game states:
+
+- The game is **ongoing**.
+- The game ended in a **draw**.
+- The game ended in a **win**.
+
+If the given board is invalid, throw an appropriate error.
+
+### Examples
+
+#### Ongoing game
+
+```
+   |   |   
+ X |   |   
+___|___|___
+   |   |   
+   | X | O 
+___|___|___
+   |   |
+ O | X |  
+   |   |
+```
+
+```julia
+julia> gamestate("X  \n XO\nOX ")
+:ongoing
+```
+
+#### Draw
+
+```
+   |   |   
+ X | O | X 
+___|___|___
+   |   |   
+ X | X | O 
+___|___|___
+   |   |
+ O | X | O
+   |   |
+```
+
+```julia
+julia> gamestate("XOX\nXXO\nOXO")
+:draw
+```
+
+#### Win
+
+```
+   |   |   
+ X | X | X 
+___|___|___
+   |   |   
+   | O | O 
+___|___|___
+   |   |
+   |   |  
+   |   |
+```
+
+```julia
+julia> gamestate("XXX\n OO\n   ")
+:win
+```
--- a/exercises/julia-1-a/README.md
+++ b/exercises/julia-1-a/README.md
@@ -38,6 +38,11 @@ There are 3 potential game states:
 
 If the given board is invalid, throw an appropriate error.
 
+If a board meets the following conditions, it is invalid:
+
+- The players took turns out of order (remember that `X` starts).
+- The game was played after it already ended, i.e. more than one row/column/diagonal contains the same symbol thrice.
+
 ### Examples
 
 #### Ongoing game
--- a/exercises/julia-1-a/README.md
+++ b/exercises/julia-1-a/README.md
@@ -1,6 +1,6 @@
 ## Instructions
 
-Implement your own [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) game.
+In this exercise, you're going to implement a program that determines the state of a [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) game.
 (_You may also know the game as "noughts and crosses" or "Xs and Os"._)
 
 The games is played on a 3x3 grid.
--- a/exercises/julia-1-a/README.md
+++ b/exercises/julia-1-a/README.md
@@ -9,7 +9,7 @@ The game ends when one player has won by placing three of marks in a row, column
 
 Without loss of generality we will assume that `X` starts.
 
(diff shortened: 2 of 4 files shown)


exercises/julia-1-a/runtests.jl
```diff
@@ -0,0 +1,214 @@
...
+        """,
+        """
+        XXX
+        OOO
+           
+        """
+    ]
+
+    for b in boards
+        @test_throws Exception gamestate(b)
```
SaschaMann, line 212: This intentionally tests for the supertype of all exceptions so that the student can choose if they want to throw an existing type of error or implement their own.

exercises/julia-1-a/runtests.jl
```diff
@@ -0,0 +1,214 @@
...
+
+@testset "Invalid boards" begin
+    boards = [
+        """
+        XX 
+           
+           
+        """,
+        """
+        OOX
```
SaschaMann, line 190: This is only an invalid state under the assumption that `X` starts which we've taken above. This is the only place where the test suite relies on that assumption to hold true, so I could also remove the test.

exercises/julia-1-a/runtests.jl
```diff
@@ -0,0 +1,214 @@
...
+
+include("tictactoe.jl")
+
+@testset "Finished games where X won via colum victory" begin
+    boards = [
+        """
+        XOO
+        X  
+        X  
+        """,
```
SaschaMann, line 11: This is parsed as follows:

```julia
"XOO\nX  \nX  \n"
```

I chose the format to make my own life a bit easier, but it has no effect on the student.

SaschaMann:
I considered adding a function `winner(board)` that returns either `"X"` or `"O"` or errors but I'm not sure if it really adds anything.

exercises/julia-1-a/README.md
```diff
@@ -0,0 +1,98 @@
+## Instructions
+
+Implement your own [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) game.
```
iHiD, line 3: I think the task is to implement a function to determine "state" of game, not create the game itself (which to me implied flow and state, etc).

Other than that, this is good! :)

exercises/julia-1-a/runtests.jl
ErikSchierboom: Excellent

exercises/julia-1-a/README.md
SaschaMann:
```suggestion
In this exercise, you will implement a computer-controlled [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) opponent.
In the first part, you will write a program that analyses the state of the game.
```

Is this better?

iHiD: No :) 

We're not telling people there **is** a second part, because that then makes them code with future requirements in mind, which invalidates the research. So all we want to mention is the state of game bit :)

SaschaMann: Ah, I didn't know that. How about this?

```suggestion
In this exercise, you're going to implement a program that determines the state of a [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) game.
```

exercises/julia-1-a/README.md
```diff
@@ -0,0 +1,98 @@
...
+(_You may also know the game as "noughts and crosses" or "Xs and Os"._)
+
+The games is played on a 3x3 grid.
+Players take turns to place `X`s and `O`s on the grid.
+The game ends when one player has won by placing three of marks in a row, column, or along a diagonal of the grid, or when the entire grid is filled up.
+
+Without loss of generality we will assume that `X` starts.
+
+At first, you will create create a function `gamestate` that takes the game board represented a string as its sole argument and returns a Symbol describing the state of the game.
```
SaschaMann, line 12:
```suggestion
Create a function `gamestate` that takes the game board represented a string as its sole argument and returns a Symbol describing the state of the game.
```

exercises/julia-1-a/.meta/example.jl
```diff
@@ -0,0 +1,36 @@
+function gamestate(bs)
```
SaschaMann, line 1: This is a rather naïve solution and probably not very performant.

SaschaMann:
Added an example solution.

I don't want to add a stub at all because it would influence the solution, for example the naming and whether someone adds type signatures or not. `function gamestate end` as very basic stub might work but I don't think most people know that this is even valid syntax.

Is this all that's required for part a or am I missing something?

exercises/julia-1-a/README.md
```diff
@@ -0,0 +1,98 @@
...
+   |   |
+```
+
+There are 3 potential game states:
+
+- The game is **ongoing**.
+- The game ended in a **draw**.
+- The game ended in a **win**.
+
+If the given board is invalid, throw an appropriate error.
```
SaschaMann, line 39: Do you think this is clear enough or should I explain what kinds of boards would be invalid?

The tests check for taking turns out of order or boards that have multiple rows/cols/diags with 3 in a row.

ErikSchierboom: It might be good to explicitly list the invalid states, as that means less thinking by the student and more coding (which is what we want).

exercises/julia-1-a/README.md
iHiD: Thanks :)

exercises/julia-1-a/README.md
```diff
@@ -0,0 +1,98 @@
...
```
SaschaMann, line 39:
```suggestion
If the given board is invalid, throw an appropriate error.

If a board meets the following conditions, it is invalid:

- The players took turns out of order (remember that `X` starts).
- The game was played after it already ended, i.e. more than one row/column/diagonal contains the same symbol thrice.
```

exercises/julia-1-a/README.md
SaschaMann: Done in 1414171

exercises/julia-1-a/README.md
```diff
@@ -0,0 +1,103 @@
+## Instructions
+
+In this exercise, you're going to implement a program that determines the state of a [tic-tac-toe](https://en.wikipedia.org/wiki/Tic-tac-toe) game.
+(_You may also know the game as "noughts and crosses" or "Xs and Os"._)
+
+The games is played on a 3x3 grid.
+Players take turns to place `X`s and `O`s on the grid.
+The game ends when one player has won by placing three of marks in a row, column, or along a diagonal of the grid, or when the entire grid is filled up.
+
+Without loss of generality we will assume that `X` starts.
```
iHiD, line 10: I don't know what this means. I'm also not sure it's needed for this part?

ErikSchierboom:
> I don't want to add a stub at all because it would influence the solution, for example the naming and whether someone adds type signatures or not. function gamestate end as very basic stub might work but I don't think most people know that this is even valid syntax.
> 
> Is this all that's required for part a or am I missing something?

For v3, the stub is there to help students get started and not have to focus on things that the exercise isn't trying to teach. For the research experiment, the stub is there mostly to help students get started. As for influencing the design, can you do something like this:

```julia
function gamestate(bs)

end
```

?

ErikSchierboom:
> I don't want to add a stub at all because it would influence the solution, for example the naming and whether someone adds type signatures or not. function gamestate end as very basic stub might work but I don't think most people know that this is even valid syntax.
> 
> Is this all that's required for part a or am I missing something?

ErikSchierboom:
Whoops. Clicked the wrong button :D

exercises/julia-1-a/.meta/example.jl
ErikSchierboom: The existing exercises all have three example solutions, to demonstrate that there are indeed three possible implementations. They might be used in the second part of the exercise too, as starting points.

SaschaMann:
> As for influencing the design, can you do something like this:

No, because whether or not a student specifies a type signature makes a difference to the solution.

`gamestate(bs)` is different to `gamestate(bs::String)` for example.

SaschaMann:
> They might be used in the second part of the exercise too, as starting points.

Will they be able to choose or is it randomized?

ErikSchierboom:
> Will they be able to choose or is it randomized?

It will definitely be random.

iHiD:
The aim is to get enough real part 1s that we don't need to use the examples but if we don't get that we might use the examples. But the student won't know either way.

ErikSchierboom:
> No, because whether or not a student specifies a type signature makes a difference to the solution.
> 
> gamestate(bs) is different to gamestate(bs::String) for example.

🤔 Well, then maybe add a comment like:

```julia
# Please implement the gamestate function
```

exercises/julia-1-a/README.md
SaschaMann: It's needed because it decides which boards are invalid due to taking turns out of order. I can reduce it to

```suggestion
For the entire exercise, we will assume that `X` starts.
```

The w.l.o.g. means that we don't have to consider games where `O` starts because the solution/implementation would be nearly identical with the only exception that you'd first replace `X`->`O` and `O`->`X`.

SaschaMann:
Added another example solution. The approach is similar in some ways (like converting it to a matrix) but checks the conditions completely different, so I still feel like it'd quality as different approach.

exercises/julia-1-a/README.md
ErikSchierboom: I also found the w.l.o.g. a bit hard to read. I like the suggested change!

SaschaMann:
> How do we go about merging this PR?

We could merge it to a julia-dev branch for now and then merge it onto master once everything is ready.

SaschaMann:
> This is looking good! I've left a comment about the stub implementation file. The .meta/config.json file also needs to be created, which can be a little bit of a chore, but is key to a good user-experience. For this, it is probably best to look at what other tracks have done in their research experiment exercises.

I'm not entirely sure how to fill out that file, even after looking at examples from the other tracks. The `cmd` key doesn't translate well because multiple commands are run in one test, and there's no reason why there shouldn't be. They test the same property, just with multiple inputs. 

Also, this assumption from the readme is wrong:

> This means that when the code does not pass one of the tests, the user won't see the code of the test that failed or the input the code was called with. To fix this, we need to provide a bit of context when we show errors to the user.

The `results.json` contains that information. Take a look at [https://github.com/exercism/julia-test-runner/blob/automated-tests/test/fixtures/one_fail/results.json](test-runner example results) for an example. The user **is** presented with the exact code that failed, what it evaluated to, and what it should have been.

An example from this exercise:

```
Ongoing games: Test Failed at ./runtests.jl:158
Expression: gamestate(b) == :ongoing
Evaluated: blah == ongoing
Stacktrace:
[...]
Test Summary: | Fail  Total
Ongoing games |    3      3
ERROR: LoadError: Some tests did not pass: 0 passed, 3 failed, 0 errored, 0 broken.
[...]
```

The only thing that's missing is the content of `b` but that doesn't matter because a) the test name gives the student all the information that's required for them to know what it looks like, and b) if they want the exact board for whatever reason, they can just print it.

**Long story short: What should be the value of `cmd`? There's no obvious value for it**

(As an aside, this is all redundant information that could be extracted from the results.json of the examples 😕)

ErikSchierboom:
Bump @iHiD for https://github.com/exercism/research_experiment_1/pull/140#pullrequestreview-453062859

ErikSchierboom:
> We could merge it to a julia-dev branch for now and then merge it onto master once everything is ready.

I'm fine with that.

SaschaMann:
I don't have write access, so you'd have to create the branch :)

> Bump @iHiD for #140 (review)

I'll add the config.json file when I have more time. We can leave this PR open until then

SaschaMann:
There it is :)

SaschaMann:
If one of you could create a julia-dev branch, I'll change the PR to point at it.

exercises/julia-1-a/.meta/config.json
```diff
@@ -0,0 +1,162 @@
+{
+    "solution_files": ["example.jl", "example2.jl"],
+    "test_file": "runtests.jl",
+    "tests": [
+        {
+            "test": "Finished game where X won via colum victory: \"XOO\nX  \nX  \"",
```
ErikSchierboom, line 6: `colum` should be `column` (there are a couple of other places with this minor typo).

PR #146, opened 2020-07-23
rust-1-a: fixed unit test according to problem specification

EQt:


ErikSchierboom:
@coriolinus Would you mind looking into this? You are the one most familiar with the exercise.

EQt:
Oh sorry. So did not know about the difference. 
So a two elements slice is **always** increasing, right?
Maybe one could elaborate on the difference in the exercise description for others that have similar problems?

coriolinus:
No, it is not the case that a two element slice is always increasing

|   | Incrementing | Increasing |
| --- | --- | --- |
| `[0, 0]` | ✘ | ✘ |
| `[0, 1]` | ✔️ | ✔️ |
| `[0, 2]` | ✘ | ✔️ |

EQt:
So an incrementing slice `s` fulfills `s[i] == s[i-1] + 1` for `i in 1..s.len()`? So it's actually only an increment of one that you seek for?

coriolinus:
Yes.

coriolinus:
WRT the definition: in programming, incrementing and decrementing do mean by 1; see https://en.wikipedia.org/wiki/Increment_and_decrement_operators.

EQt:
Ok. Thanks

ErikSchierboom:
It might be worth considering if we need to update the instructions a little bit. I'll work on that!

EQt:
:thumbsup:

iHiD:
@ErikSchierboom Yeah - if you could add a note to explain incrementing please. Maybe something like "... incrementing (to increase by 1) ..."

PR #164, opened 2021-02-10
typo: "NINER" -> "NINE"

stragu:
Minor typo, bu not sure if this has implications for the data collection...
At least, it is not a part of the dictionary that is tested when validating the answer.

### DIFF
--- a/exercises/python-1-b/python_1_b.py
+++ b/exercises/python-1-b/python_1_b.py
@@ -40,7 +40,7 @@ ALPHANUM_TO_NATO = {
     "6": "SIX",
     "7": "SEVEN",
     "8": "EIGHT",
-    "9": "NINER",
+    "9": "NINE",
 }
 
 


exercises/python-1-b/python_1_b.py
```diff
@@ -40,7 +40,7 @@
     "6": "SIX",
     "7": "SEVEN",
     "8": "EIGHT",
-    "9": "NINER",
+    "9": "NINE",
```
kotp, line 43: Not a typographical error.  It is indeed "NINER".

kotp: Reasons:  ATC (air traffic control), military communications, and international communication where radio noise can be significant, it is important to not confuse words. I had to scroll up and make sure that "oscar" and/or "mike" where used.  Technically speaking, we were trained to say "fife" instead of the softer sounding "v" in "five" and similar clarifying sounds where communication mistakes can cost lives.

The NATO alphabet including numbers can be found [here](https://en.wikipedia.org/wiki/NATO_phonetic_alphabet) adopted by others, including the ICAO (International Civil Aviation Organization).

stragu: Right, I did look at the WIkipedia article but only looked at the "Telephony"/"Code Word" columns.
Thanks for the explanation!

kotp: Can you close PR if you think it is good as is.

